PR #10, opened 2018-07-18
Added online emulator

knoxknox:
This commit adds online asm emulator.
It includes a subset of x86 assembler instructions to learn.
More details about how this works can be found here: http://www.emu86.org/help

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 asm
 ==============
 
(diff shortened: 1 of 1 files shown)


knoxknox:
Hi @0xAX,
What do you think if we include online emulator with visualization ?

PR #13, opened 2020-02-13
fix sys_write to 4

plutowang:


0xAX:
Hello @plutowang. What is the reason for this change? `sys_write` number is `1` for Linux x86_64

PR #29, opened 2024-06-16
Rework the second chapter

0xAX:


content/assets/stack-preserve-bp.svg
klaudiagrz, line 1: Please unify to start with capital letters

content/assets/stack.svg
klaudiagrz, line 1: Please unify to start with capital letters

sum/README.md
```diff
@@ -1,8 +1,11 @@
-This is "sum of 2 numbers" with assembly for Linux x86-64
+# Sum of numbers
 
-Build it with: `make`
+This is a basic program that gets the sum of two integer numbers and check that it is correct. It prints the message if the sum is correct, otherwise just exits.
```
klaudiagrz, line 3:
```suggestion
This is a basic program that sums two integer numbers and checks if it is correct. If the sum is correct, it prints the message; otherwise, it just exits.
```

sum/README.md
```diff
@@ -1,8 +1,11 @@
...
 
-More details at - [Say hello to x64 Assembly [part 2]](https://github.com/0xAX/asm/blob/master/content/asm_2.md)
+To build the program, run:
 
-[@0xAX](https://x.com/0xAX)
+```bash
+make
+```
 
+For more details, read [Part 2. The `x86_64` concepts](https://github.com/0xAX/asm/blob/master/content/asm_2.md)
```
klaudiagrz, line 11:
```suggestion
For more details, read [Part 2. The `x86_64` concepts](https://github.com/0xAX/asm/blob/master/content/asm_2.md).
```

klaudiagrz (commented):
Comments to README.md

sum/sum.asm
```diff
@@ -1,47 +1,53 @@
...
-	mov	rax, num1
-	mov	rbx, num2
-	add	rax, rbx
-	; compare rax with correct sum - 150
-	cmp	rax, 150
-	; if rax is not 150 go to exit
-	jne	.exit
-	; if rax is 150 print msg
-	jmp	.rightSum
+    ;; Set the value of the num1 to the rax
```
klaudiagrz, line 17:
```suggestion
    ;; Set the value of num1 to rax
```

klaudiagrz: No need for articles here. Please adjust also in the descriptions below.

sum/sum.asm
```diff
@@ -1,47 +1,53 @@
...
+    mov rax, [num1]
+    ;; Set the value of the num2 to the rbx
+    mov rbx, [num2]
+    ;; Get sum of the rax and rbx. The result is stored in the rax.
+    add rax, rbx
+.compare:
+    ;; Compare the value of the rax with `150`
```
klaudiagrz, line 24:
```suggestion
    ;; Compare the rax value with 150
```

klaudiagrz: In other descriptions, you are not using `code`. Here also no need.

sum/sum.asm
```diff
@@ -1,47 +1,53 @@
...
+    cmp rax, 150
+    ;; Go to the .exit label if the values of the rax and 150 are not equal
```
klaudiagrz, line 26:
```suggestion
    ;; Go to the .exit label if the rax value is not 150
```

sum/sum.asm
```diff
@@ -1,47 +1,53 @@
...
+    jne .exit
+    ;; Go to the .correctSum label if the values of the rax and 150 are equal
```
klaudiagrz, line 28:
```suggestion
    ;; Go to the .correctSum label if the rax value is 150
```

sum/sum.asm
```diff
@@ -1,47 +1,53 @@
...
-	mov	rsi, msg
-	;; length of message
-	mov	rdx, 15
-	;; call write syscall
-	syscall
-	; exit from program
-	jmp	.exit
+; Print message that the sum is correct
+.correctSum:
+    ;; Number of the sytem call. 1 - `sys_write`.
```
klaudiagrz, line 33:
```suggestion
    ;; Specify the system call number (1 is `sys_write`).
```

klaudiagrz: Maybe we can unify with the descriptions from `hello.asm`?

klaudiagrz (commented):
Comments to sum.asm

content/assets/stack-preserve-bp.svg
0xAX: Fixed here and in all other diagrams

klaudiagrz (commented):
Review of the blog post till the "Stack" section

sum/sum.asm
```diff
@@ -1,47 +1,53 @@
...
-	add	rax, rbx
-	; compare rax with correct sum - 150
-	cmp	rax, 150
-	; if rax is not 150 go to exit
-	jne	.exit
-	; if rax is 150 print msg
-	jmp	.rightSum
+    ;; Set the value of num1 to rax
+    mov rax, [num1]
+    ;; Set the value of the num2 to the rbx
```
klaudiagrz, line 19:
```suggestion
    ;; Set the value of num2 to rbx
```

sum/sum.asm
```diff
@@ -1,47 +1,53 @@
...
+    mov rbx, [num2]
+    ;; Get sum of the rax and rbx. The result is stored in the rax.
```
klaudiagrz, line 21:
```suggestion
    ;; Get the sum of rax and rbx. The result is stored in rax.
```

sum/sum.asm
```diff
@@ -1,47 +1,53 @@
...
 
 ; exit procedure
 .exit:
-	mov	rax, 60
-	mov	rdi, 0
-	syscall
\ No newline at end of file
+    ;; Specify the number of the system call (60 is `sys_exit`).
+    mov rax, 60
+    ;; Set the first argument of `sys_exit` to `0`. The `0` status code is success.
```
klaudiagrz, line 50:
```suggestion
    ;; Set the first argument of `sys_exit` to 0. The 0 status code is success.
```

klaudiagrz (commented):
More comments to sum.asm

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+For example:
 
-Ok, it is almost all clear here. 3 constants with name num1, num2, msg and with values 100, 50 and "Sum is correct", 10. But what is it db, equ? Actual NASM supports a number of pseudo-instructions:
+```assembly
+section .bss
+    ;; Define a buffer with the size 64 bytes
+    buffer resb 64
+```
+
+After variables are defined we can start to use them in the code of our program. To use a variable we can refer it by name. Although there is a small thing to remember in the NASM assembly syntax. If we will access a variable just by a name, we will get the address of it but not the actual value this variable stores:
```
klaudiagrz, line 191:
```suggestion
After defining variables, we can start using them in our program's code. To use a variable, we can refer it by name. However, there is a small thing to remember in the NASM assembly syntax: accessing a variable by name gets us its address, not the actual value it stores:
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+
+```assembly
+;; Move the address of the `num1` variable to the al register
+move al, num1
+```
 
-* DB, DW, DD, DQ, DT, DO, DY and DZ - are used for declaring initialized data. For example:
+If we want to get the actual value that is located by the given address we need to specify the variable name in square brackets:
```
klaudiagrz, line 198:
```suggestion
To get the actual value that located in the given address, we need to specify the variable name in square brackets:
```

klaudiagrz (commented):
Re-check till "Stack"

content/asm_2.md
```diff
@@ -1,235 +1,510 @@
...
+
+To do the calculation we need to access the input parameters. As you may see it is done using the address stored in the `rbp` register and offsets from it. The offsets are negative as you may remember the stack grows down towards lower addresses. At first we move the value stored in the `edi` register (the first argument of the `foo` function) to address stored in the `rbp` register with the `-4` (the offset is negative because you should remember that stack grows down) bytes offset. After that we move the value stored in the `esi` register (the second argument of the `foo` function) to the address stored in the `rbp` register with the `-8` bytes offset. We repeat these operations for the all six input arguments.
+
+Now take a look one more time very carefully:
+
+> to the address stored in the `rbp` register with the `N` byte offset
+
+What was the address stored in the `rbp`? Our stack pointer! So after the last `mov` instruction in the function `foo`, our stack frame will look like this:
```
klaudiagrz, line 343:
```suggestion
What was the address stored in the `rbp` register? Our stack pointer! So after the last `mov` instruction in the function `foo`, our stack frame will look like this:
```

content/asm_2.md
```diff
@@ -1,235 +1,510 @@
...
+
+![stack](./assets/stack.svg)
+
+That is the whole sense of the `rbp`. It plays role of an anchor in the function or a base point. Using the positive offsets we may access the return address and the parameters pushed on the stack by the caller and using the negative offsets we may access local variables.
```
klaudiagrz, line 347:
```suggestion
That is the whole point of the `rbp` register. It plays the role of an anchor or a base point in the function. Using the positive offsets, we can access the return address and parameters pushed onto the stack by the caller. On the other hand, negative offsets allow us to access local variables of the current function.
```

content/asm_2.md
```diff
@@ -1,235 +1,510 @@
...
+
+Right before the return from the `foo` function we may see so called [function epilogue](https://en.wikipedia.org/wiki/Function_prologue_and_epilogue#Epilogue) we restore the initial value of the `rbp` by removing it from the stack. The last `ret` instruction pops the return address from the stack and the execution continues from this address.
```
klaudiagrz, line 349:
```suggestion
Right before returning from the `foo` function, we encounter the so-called [function epilogue](https://en.wikipedia.org/wiki/Function_prologue_and_epilogue#Epilogue). During this step, we restore the initial value of the `rbp` by removing it from the stack. Finally, the last `ret` instruction pops the return address from the stack, and the program's execution continues from this address.
```

content/asm_2.md
```diff
@@ -1,235 +1,510 @@
...
+
+- `push` - push the operand on the stack.
+- `pop` - pop the top value from the stack.
+
+The stack grows downwards from the high addresses to low. So, basically when we hear `top of the stack`, it means the lowest address. The general purpose registers `rsp` always should point to the top of the stack. In the [system call](#system-call) section, we have seen that first six arguments of a system call are passed in the general purpose registers. According to the calling conventions document:
```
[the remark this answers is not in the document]
0xAX: yes, especially technically it could be not always... so I replaced with `should point`

content/asm_2.md
```diff
@@ -1,235 +1,510 @@
...
 
 ```assembly
-;; now one is 1
-one equ 1
+push    rbp
+mov     rbp, rsp
 ```
 
-* TIMES - Repeating Instructions or Data. (description will be in next posts)
+The names of these two instructions in the beginning of each function is - [function prologue](https://en.wikipedia.org/wiki/Function_prologue_and_epilogue#Prologue). Each function usually operates with a part of the stack. Such part is called a [stack frame](https://en.wikipedia.org/wiki/Call_stack). To manage stack CPU is using the several general purpose registers:
```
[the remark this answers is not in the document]
0xAX: `function`, so this is correct

content/asm_2.md
```diff
@@ -1,235 +1,510 @@
...
+int bar() {
+    return foo(1, 2, 3, 4, 5, 6, 7, 8);
+}
+```
 
-;; Initialize word to 0x12 0x34
-dw    0x1234
+If we will compile it and have a look at the assembly code, we will see something like that:
+
+```assembly
```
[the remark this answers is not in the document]
0xAX: Fixed as discussed

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+        mov     ecx, 4
+        ;; Push the third argument on the stack
+        mov     edx, 3
+        ;; Push the second argument on the stack
+        mov     esi, 2
+        ;; Push the first argument on the stack
+        mov     edi, 1
+        ;; Call the function `foo`
+        call    foo
+        ;; Clean-up the stack from the 8th and 7th arguments
```
klaudiagrz, line 271:
```suggestion
        ;; Clean up the stack from the 8th and 7th arguments
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+        ... # Skip arithmetic operations for now
+        ...
+        ;; Restore the old rbp
+        pop     rbp
+        ;; Return from the function
+        ret
+```
+
+> [!NOTE]
+> The C program should be compiled without any optimization flags because the compiler may eliminate all the calculation and just calculate the sum of arguments in compile time. You can use `-O0 -masm=intel` flags for compiler to avoid optimization. You can use tools like [godbolt](https://godbolt.org/) to see the assembly output of these functions.
```
klaudiagrz, line 304:
```suggestion
> The C program should be compiled without any optimization flags because the compiler may eliminate all the calculations and just calculate the sum of arguments in compile time. You can use `-O0 -masm=intel` flags for the compiler to avoid optimization. You can use tools like [godbolt](https://godbolt.org/) to see the assembly output of these functions.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+
+- `rip` - the general purpose register `rip` is the so-called `instruction pointer`. This register stores the address of the next instruction the CPU is going to execute. When the CPU meets the `call` instruction to call a function, it pushes the address of the next instruction to run after the function call to the stack. This is done so the CPU knows where to continue the program's execution after the function call.
+- `rsp` - the `rsp` register is called a `stack pointer` and should point to the top of the stack. After we push something to the stack using the `push` instruction, the stack pointer address decreases. After we pop something from the stack using the `pop` instruction, the stack pointer address increases.
+- `rbp` - the general purpose register `rbp` is the so-called `frame pointer` or `base pointer` that points to the stack frame. As mentioned above, each function has its own stack frame, which is a memory area where the function stores [local variables](https://en.wikipedia.org/wiki/Local_variable) and other data.
```
klaudiagrz, line 317:
```suggestion
- `rip` - this register is the so-called `instruction pointer`. It stores the address of the next instruction the CPU is going to execute. When the CPU meets the `call` instruction to call a function, it pushes the address of the next instruction to run after the function call to the stack. This is done so the CPU knows where to continue the program's execution after the function call.
- `rsp` - this register is called a `stack pointer` and should point to the top of the stack. After we push something to the stack using the `push` instruction, the stack pointer address decreases. After we pop something from the stack using the `pop` instruction, the stack pointer address increases.
- `rbp` - this register is the so-called `frame pointer` or `base pointer` that points to the stack frame. As mentioned above, each function has its own stack frame, which is a memory area where the function stores [local variables](https://en.wikipedia.org/wiki/Local_variable) and other data.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+
+For calculations, we need to access the input parameters. This is done using the address stored in the `rbp` register and offsets from this address. These offsets are negative because the stack grows down towards lower memory addresses. At first, we move the value stored in the `edi` register (the first argument of the `foo` function) to the address stored in the `rbp` register with the `-4` bytes offset. Similarly, we move the value stored in the `esi` register (the second argument) to the address stored in the `rbp` register with the `-8` bytes offset. We repeat this operation for all six input arguments.
+
+Let's read once again the sentence from the paragraph above:
+
+> For calculations, we need to access the input parameters. This is done using the address stored in the `rbp` register and offsets from this address.
+
+What was the address stored in the `rbp` register? Our stack pointer! So after the last `mov` instruction in the function `foo`, our stack frame will look like this:
```
klaudiagrz, line 341:
```suggestion
What was the address stored in the `rbp` register? Our stack pointer! So after running the last `mov` instruction in the function `foo`, our stack frame looks like this:
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+
+![stack](./assets/stack.svg)
+
+That is the whole point of the `rbp` register. It plays the role of an anchor or a base point in the function. Using the positive offsets, we can access the return address and parameters pushed onto the stack by the caller. On the other hand, negative offsets allow us to access local variables of the current function.
+
+Right before returning from the `foo` function, we encounter the so-called [function epilogue](https://en.wikipedia.org/wiki/Function_prologue_and_epilogue#Epilogue). During this step, we restore the initial value of the `rbp` by removing it from the stack. Finally, the last `ret` instruction pops the return address from the stack, and the program's execution continues from this address.
```
klaudiagrz, line 347:
```suggestion
Right before returning from the `foo` function, we encounter the so-called [function epilogue](https://en.wikipedia.org/wiki/Function_prologue_and_epilogue#Epilogue). During this step, we restore the initial value of `rbp` by removing it from the stack. Finally, the last `ret` instruction pops the return address from the stack, and the program's execution continues from this address.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
 ```
 
-* RESB, RESW, RESD, RESQ, REST, RESO, RESY and RESZ - are used for declaring non initialized variables
-* INCBIN - includes External Binary Files
-* EQU - defines constant. For example:
+### Stack
+
+We can not dive into assembly programming without understanding one of the crucial concepts of the `x86_64` (and other) architectures — the [stack](https://en.wikipedia.org/wiki/Stack_(abstract_data_type). The stack is a specific memory area in a program that is accessed in a last-in, first-out (LIFO) pattern.
+
+A processor has a very limited number of registers. As we already know, an `x86_64` processor provides access to only 16 general purpose registers. This number is very limited. Often, we may need more – or even much more space – to store data. One way to solve this issue is by using the program's stack. We can look at the stack as the usual memory area but with one significant difference – its access pattern.
```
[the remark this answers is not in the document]
klaudiagrz: repetition of the first sentence

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+So the available number of general purpose registers should be enough to execute any system call. But what about other functions? What if one has more than six arguments? In this case, the first six parameters are also passed in general purpose registers and all the next parameters are passed on the stack. The set of general purpose registers to call a library function is slightly different from the set of registers used for a system call:
+
+- `rdi` - used to pass the first argument to a function.
+- `rsi` - used to pass the second argument to a function.
+- `rdx` - used to pass the third argument to a function.
+- `rcx` - used to pass the fourth argument to a function.
+- `r8` - used to pass the fifth argument to a function.
+- `r9` - used to pass the sixth argument to a function.
+
+Let’s take a look at the assembly code of a simple example written in the C programming language to better illustrate the topic:
```
klaudiagrz, line 233:
```suggestion
Let’s take a look at the simple code example written in the C programming language to better illustrate the topic:
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
-;; now one is 1
-one equ 1
+push    rbp
+mov     rbp, rsp
 ```
 
-* TIMES - Repeating Instructions or Data. (description will be in next posts)
+These two instructions at the beginning of each function are called [function prologue](https://en.wikipedia.org/wiki/Function_prologue_and_epilogue#Prologue). Each function usually operates with a part of the stack. Such a part is called a [stack frame](https://en.wikipedia.org/wiki/Call_stack). To manage the stack, the CPU uses these general purpose registers:
+
+- `rip` - this register `rip` is the so-called `instruction pointer`. This register stores the address of the next instruction the CPU is going to execute. When the CPU meets the `call` instruction to call a function, it pushes the address of the next instruction to run after the function call to the stack. This is done so the CPU knows where to continue the program's execution after the function call.
```
klaudiagrz, line 315:
```suggestion
- `rip` - this register is the so-called `instruction pointer`. It stores the address of the next instruction the CPU is going to execute. When the CPU meets the `call` instruction to call a function, it pushes the address of the next instruction to run after the function call to the stack. This is done so the CPU knows where to continue the program's execution after the function call.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+- `rsp` - this register is called a `stack pointer` and should point to the top of the stack. After we push something to the stack using the `push` instruction, the stack pointer address decreases. After we pop something from the stack using the `pop` instruction, the stack pointer address increases.
+- `rbp` - this register `rbp` is the so-called `frame pointer` or `base pointer` that points to the stack frame. As mentioned above, each function has its own stack frame, which is a memory area where the function stores [local variables](https://en.wikipedia.org/wiki/Local_variable) and other data.
```
klaudiagrz, line 317:
```suggestion
- `rbp` - this register is the so-called `frame pointer` or `base pointer` that points to the stack frame. As mentioned above, each function has its own stack frame, which is a memory area where the function stores [local variables](https://en.wikipedia.org/wiki/Local_variable) and other data.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
-    ;; ....
-    jmp .exit
+### Program example
 
-.exit:
-    mov    rax, 60
-    mov    rdi, 0
-    syscall
-```
+Since we learned some basic arithmetic and control flow instructions, we can write our example. Before we will take a look at the program source code, I will remind that we are going to write a simple program that will calculate the sum of two integer numbers and if the sum equal to the third predefined number we will print a string. Otherwise just exist.
```
klaudiagrz, line 422:
```suggestion
Now that we've covered basic arithmetic and control flow instructions, we can write a simple program. As a reminder, this program will calculate the sum of two integers and compare it to a predefined third number. If the sum equals the third number, a string will be printed; otherwise, the program will just exit.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
-    mov     rdi, 1
-    ;; message address
-    mov     rsi, msg
-    ;; length of message
-    mov     rdx, 15
-    ;; call write syscall
+    ;; Go to the .correctSum label if the rax value is 150
+    jmp .correctSum
+
+; Print message that the sum is correct
```
klaudiagrz, line 457:
```suggestion
;; Print a message that the sum is correct
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+    ;; Set the first argument of `sys_write` to 1 (`stdout`).
+    mov rdi, 1
+    ;; Set the second argument of `sys_write` to the reference of the `msg` variable.        
+    mov rsi, msg
+    ;; Set the third argument to the length of the `msg` variable's value (20 bytes).
+    mov rdx, 20
+    ;; Call the `sys_write` system call.
     syscall
-    ; exit from program
+    ; Go to the exit of the program.
```
klaudiagrz, line 469:
```suggestion
    ;; Go to the exit of the program.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
     jmp .exit
 
 ; exit procedure
```
klaudiagrz, line 472:
```suggestion
;; Exit procedure
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+    ;; Specify the number of the system call (60 is `sys_exit`).
+    mov rax, 60
+    ;; Set the first argument of `sys_exit` to 0. The 0 status code is success.
+    mov rdi, 0
+    ;; Call the `sys_exit` system call.
     syscall
 ```
 
-Let's go through the source code. First of all there is data section with two constants num1, num2 and variable msg with "Sum is correct\n" value. Now look at 14 line. There is begin of program's entry point. We transfer num1 and num2 values to general purpose registers rax and rbx. Sum it with add instruction. After execution of add instruction, it calculates sum of values from rax and rbx and store it's value to rax. Now we have sum of num1 and num2 in the rax register.
+First of all let's try to build, run our program with the similar commands that we have seen in the previous chapter and see the result:
```
klaudiagrz, line 482:
```suggestion
First, let's build and run our program using the commands we learned in the previous chapter to see the result:
```

klaudiagrz: What do these commands do? Do they really both build and run the program? Or only build? Because below we have the command that actually runs the program:

> After we built our program, we can run it with:

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+
+```bash
+$ nasm -f elf64 -o program.o program.asm
+$ ld -o program program.o
+```
+
+After we built our program, we can run it with:
```
klaudiagrz, line 489:
```suggestion
After building the program, we can run it with:
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+
+```bash
+~$ ./program
+Sum is correct
+```
+
+Now let's go through the source code of our program. First of all there is the `.data` section with three variables:
```
klaudiagrz, line 496:
```suggestion
Now let's go through the source code of our program. First of all, there is the `.data` section with three variables:
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+
+- `num1`
+- `num2`
+- `msg`
+
+The entry point of our program is the `_start` symbol. In the beginning of the source code of our program we put the values of the `num1` and `num2` to the general purpose registers `rax` and `rbx`. After this we can use the `add` instruction to get the sum of these two values. The result of the sum will be stored in the `rax` register.
```
klaudiagrz, line 502:
```suggestion
The entry point of our program is the `_start` symbol. At the beginning of the source code, we put the values of `num1` and `num2` to the general purpose registers `rax` and `rbx`. Then, we use the `add` instruction to calculate their sum, which will be stored in the `rax` register.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+
+According to the description of our program, now we have to compare the sum of two numbers with the predefined number. We do it with the `cmp` instruction. At this point we have two ways to go. The first one - we jump to the `.exit` label if the value of the `rax` (that stores sum of the `num1` and `num2` values) is not equal to `150`. If the sum is equal to `150`, we jump to the `.correctSum` label.
```
klaudiagrz, line 504:
```suggestion
Based on our program's description, the next step is to compare the sum of the two numbers with the predefined value. We use the `cmp` instruction for this. At this point, there are two possible outcomes:

- If the value in the `rax` register (which stores the sum of `num1` and `num2`) is not equal to 150, we jump to the `.exit` label.
- If the sum is equal to 150, we jump to the `.correctSum` label.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
 
-Ok we have num1 which is 100 and num2 which is 50. Our sum must be 150. Let's check it with cmp instruction. After comparison rax and 150 we check result of comparison, if rax and 150 are not equal (checking it with jne) we go to .exit label, if they are equal we go to .rightSum label.
+The source code of the both `.correctSum` and `.exit` sub-routines should be familiar to us. They both do very similar what we already have seen in the previous chapter. The `.correctSum` sub-routine prints the string from the `msg` to the screen. The `.exit` sub-routine provides us graceful exit from our program.
```
klaudiagrz, line 506:
```suggestion
The source code of both `.correctSum` and `.exit` sub-routines should look familiar, as they are similar to what we covered in the previous chapter. The `.correctSum` sub-routine prints the string from `msg` to the screen, while the `.exit` sub-routine ensures a graceful exit from the program.
```

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
 
-Now we have two labels: .exit and .rightSum. First is just sets 60 to rax, it is exit system call number, and 0 to rdi, it is a exit code. Second is .rightSum is pretty easy, it just prints Sum is correct.
+That is it for this post. In the next post we will continue to dive into assembly programming.
```
klaudiagrz, line 508:
```suggestion
We’ve just written our second program — great job! In the next post, we’ll continue exploring assembly programming. If you have any questions or thoughts, feel free to reach out. See you in the next post!
```

klaudiagrz: Not sure if it's not too cheerful, though 🤔

0xAX: No, very good

content/asm_2.md
0xAX: Yes, nice catch. Removed 'run' from the original sentence.

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
-*  `JA` - the same that JG, but performs unsigned comparison
-*  `JAE` - the same that JGE, but performs unsigned comparison
+-  `JE`  - Jump if the values are equal.
+-  `JNE` - Jump if the values are not equal.
+-  `JZ`  - Jump if the difference between the two values is zero.
+-  `JNZ` - Jump if the difference between the two values is not zero.
+-  `JG`  - Jump if the first value is greater than the second.
+-  `JGE` - Jump if the first is greater or equal to the second.
+-  `JA`  - The same that JG, but performs unsigned comparison.
+-  `JAE` - The same that JGE, but performs unsigned comparison.
```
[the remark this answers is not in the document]
0xAX: Actually the same. Very good catch. I combined them into one `JE/JZ`

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
-* `SUB` - substract
-* `MUL` - unsigned multiply
-* `IMUL` - signed multiply
-* `DIV` - unsigned divide
-* `IDIV` - signed divide
-* `INC` - increment
-* `DEC` - decrement
-* `NEG` - negate
+- `ADD`  - Addition.
+- `SUB`  - Substraction.
```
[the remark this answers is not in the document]
0xAX: Removed.

content/asm_2.md
```diff
@@ -1,235 +1,508 @@
...
+
+Right before returning from the `foo` function, we encounter the so-called [function epilogue](https://en.wikipedia.org/wiki/Function_prologue_and_epilogue#Epilogue). During this step, we restore the initial value of `rbp` by removing it from the stack. Finally, the last `ret` instruction pops the return address from the stack, and the program's execution continues from this address.
+
+## Example
+
+After we went thought the most important concepts, it is time to return to the most interesting part - writing the programs. Let's take a look at our second simple assembly program. The program will take two integer numbers, get the sum of these numbers and compare it with the third predefined number. If the predefined number is equal to sum, the program will print something on the screen, if not - the program will just exit.
+
+Before writing the code we need to know how to execute basic arithmetic expressions and compare the things.
```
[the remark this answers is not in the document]
0xAX: I reworked it a bit, please take a look.

content/asm_2.md
```diff
@@ -1,235 +1,510 @@
...
-* `NEG` - negate
+- `ADD`  - Addition
+- `SUB`  - Subtraction
+- `MUL`  - Unsigned multiplication
+- `IMUL` - Signed multiplication
+- `DIV`  - Unsigned division
+- `IDIV` - Signed division
+- `INC`  - Increment
+- `DEC`  - Decrement
+- `NEG`  - Negate
```
klaudiagrz, line 367:
```suggestion
- `NEG`  - Negation
```

content/asm_2.md
```diff
@@ -1,235 +1,510 @@
...
+- If the sum is equal to 150, we jump to the `.correctSum` label.
 
-Ok we have num1 which is 100 and num2 which is 50. Our sum must be 150. Let's check it with cmp instruction. After comparison rax and 150 we check result of comparison, if rax and 150 are not equal (checking it with jne) we go to .exit label, if they are equal we go to .rightSum label.
+The source code of both `.correctSum` and `.exit` sub-routines should look familiar, as they are similar to what we covered in the previous chapter. The `.correctSum` sub-routine prints the string from `msg` to the screen, while the `.exit` sub-routine ensures a graceful exit from the program.
 
-Now we have two labels: .exit and .rightSum. First is just sets 60 to rax, it is exit system call number, and 0 to rdi, it is a exit code. Second is .rightSum is pretty easy, it just prints Sum is correct.
+We’ve just written our second program — great job 🎉 In the next post, we’ll continue exploring assembly programming and dive deeper into the stack topic. If you have any questions or thoughts, feel free to reach out. See you in the next post!
```
klaudiagrz, line 510: :)

klaudiagrz (approved):
Looks good!

content/asm_2.md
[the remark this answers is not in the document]
klaudiagrz: Shouldn't be the same with JNE and JNZ?

content/asm_2.md
```diff
@@ -1,235 +1,510 @@
...
-* `NEG` - negate
+- `ADD`  - Addition
+- `SUB`  - Subtraction
+- `MUL`  - Unsigned multiplication
+- `IMUL` - Signed multiplication
+- `DIV`  - Unsigned division
+- `IDIV` - Signed division
+- `INC`  - Increment
+- `DEC`  - Decrement
+- `NEG`  - Negation.
```
klaudiagrz, line 367:
```suggestion
- `NEG`  - Negation
```

content/asm_2.md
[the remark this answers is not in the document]
0xAX: Yes, you are right. Also combined them into one line.

PR #41, opened 2025-01-07
Rework third chapter

0xAX:
**Description**

This PR provides rework of the third chapter.


stack/stack.asm
```diff
@@ -1,168 +1,161 @@
...
-	mul	r12
-	mov	rdx, rax
-	;;;;
-
-	;;;; print sum
-	mov	rax, SYS_WRITE
-	mov	rdi, STD_IN
-	mov	rsi, rsp
-	;; call sys_write
+;; Convert the sum to string and print on the screen.
```
klaudiagrz, line 102:
```suggestion
;; Convert the sum to string and print it on the screen.
```

stack/stack.asm
```diff
@@ -1,168 +1,161 @@
...
+int_to_str:
+	;; High part of dividend. The low part is in the rax register.
+	mov rdx, 0
+	;; Set the divisor to 10.
+	mov rbx, 10
+	;; Divide the sum (rax from rax) to 10. Reminder will be stored in the rdx register.
+	div rbx
+	;; Add 48 to the reminder to get string ASCII representation of the number value.
```
klaudiagrz, line 110:
```suggestion
	;; Add 48 to the reminder to get a string ASCII representation of the number value.
```

stack/stack.asm
```diff
@@ -1,168 +1,161 @@
...
+	add rdx, 48
+	;; Store reminder on the stack.
```
klaudiagrz, line 112:
```suggestion
	;; Store the reminder on the stack.
```

stack/stack.asm
```diff
@@ -1,168 +1,161 @@
...
+	;; Otherwise print the result.
+	jmp printResult
+
+;; Print result to the standard output.
+printResult:
+	;; Put the number of symbols within the string to the rax register.
+        mov rax, rcx
+        ;; Put the value 8 to the rcx register.
+        mov rcx, 8
+        ;; Calculate the number of bytes in the given string by multiplication rax to 8.
```
klaudiagrz, line 129:
```suggestion
        ;; Calculate the number of bytes in the given string by multiplying rax by 8.
```

stack/stack.asm
```diff
@@ -1,168 +1,161 @@
...
+	;; Increase the counter.
+	inc rcx
+	;; Compare the rest of the sum with zero.
+	cmp rax, 0x0
+	;; If it is not zero yet, continue to convert it to string.
+	jne int_to_str
+	;; Otherwise print the result.
+	jmp printResult
+
+;; Print result to the standard output.
```
klaudiagrz, line 123:
```suggestion
;; Print the result to the standard output.
```

stack/stack.asm
```diff
@@ -1,168 +1,161 @@
...
-;;
 exit:
-	;; syscall number
-	mov	rax, SYS_EXIT
-	;; exit code
-	mov	rdi, EXIT_CODE
-	;; call sys_exit
+	;; Specify the number of the system call (60 is `sys_exit`).
+	mov rax, SYS_EXIT
+	;; Set the first argument of `sys_exit` to 0. The 0 status code is success.
```
klaudiagrz, line 158:
```suggestion
	;; Set the first argument of `sys_exit` to 0. The 0 status code is a success.
```

klaudiagrz (commented):
Review of all files except the main post

klaudiagrz:
Please adjust the descriptions also in the code sample of the third post.

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
 
-We have two commands for work with stack:
+The third instruction of the function `__double` puts the first parameter of this function to the stack with offset `-20`. After this we may see that the value `2` which is the value of the local variable `two` is also put onto the stack with the offset `-4`. The stack frame of our function for this moment should look like this: 
 
-* `push argument` - increments stack pointer (RSP) and stores argument in location pointed by stack pointer
-* `pop argument` - copied data to argument from location pointed by stack pointer
+![asm-3-stack-fram-of__double-2](./assets/asm-3-stack-of__double-2.svg)
 
-Let's look on one simple example:
+After this we put the value from the stack with the offset `-20` (the value of the function's parameter) to the register eax and multiply it by `2` which is located on the stack with the offset `-4`. The result of the multiplication will be in the register eax. This simple example shows how stack is used to access and parameters and local variables of the function.
```
klaudiagrz, line 89:
```suggestion
Finally, we put the value from the stack at offset `-20` (the value of the function's parameter) into the `eax` register and multiply it by `2`, which is located on the stack at offset `-4`. The result of the multiplication is then stored in the `eax` register. This simple example shows how the stack is used to access both parameters and local variables of a function.
```

klaudiagrz (commented):
Review of the intro section

stack/stack.asm
```diff
@@ -1,168 +1,161 @@
...
+	mov rcx,  10
+__repeat:
+	;; Compare the first element in the given string with the NUL terminator (end of string).
+	cmp [rsi], byte 0
+	;; If we reached the end of the string, return from the procedure. The result is stored in the rax register.
+	je __return
+	;; Move the current character from the command line argument to the bl register.
+	mov bl, [rsi]
+	;; Subtract the value 48 from the ASCII code of the current character.
+        ;; This will give us numeric value of the character.
```
klaudiagrz, line 88:
```suggestion
        ;; This will give us the numeric value of the character.
```

stack/stack.asm
```diff
@@ -1,168 +1,161 @@
...
+	sub bl, 48
+	;; Multiple our result number by 10 to get the place for the next digit.
+	mul rcx
+	;; Add the next digit to our result number.
+	add rax, rbx
+	;; Move to the next character in the command line argument string.
+	inc rsi
+	;; Repeat while we did not reach the end of string.
```
klaudiagrz, line 96:
```suggestion
	;; Repeat until we do not reach the end of the string.
```

stack/stack.asm
```diff
@@ -1,168 +1,161 @@
...
-	;;;;
-
-	;;;; print sum
-	mov	rax, SYS_WRITE
-	mov	rdi, STD_IN
-	mov	rsi, rsp
-	;; call sys_write
+;; Convert the sum to string and print it on the screen.
+int_to_str:
+	;; High part of dividend. The low part is in the rax register.
```
klaudiagrz, line 104:
```suggestion
	;; High part of the dividend. The low part is in the rax register.
```

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
+- `r8` - used to pass the fifth argument to a function.
+- `r9` - used to pass the sixth argument to a function.
+
+Local variables are also accessed using the stack. For example, let's take a look at the following C function that doubles its parameter:
 
 ```C
-int foo(int a1, int a2, int a3, int a4, int a5, int a6, int a7)
-{
-    return (a1 + a2 - a3 - a4 + a5 - a6) * a7;
+// The __ prefix in the __dobule function name to not mix it with the `double` data type.
```
klaudiagrz, line 52:
```suggestion
// The "__" prefix in the `__double` function name is used to avoid confusion with the `double` data type.
```

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
+If we compile this function and take a look at the assembly output, we will see something like this:
 
-## Stack pointer
+```assembly
+__double(int):
+        ;; Preserve the base pointer
+        push    rbp
+        ;; Set the new frame base pointer
+        mov     rbp, rsp
+        ;; Put the value of the first parameter of the function from the edi register on the stack with the location rbp - 20 bytes.
```
klaudiagrz, line 68: The GitHub preview cuts the sentence after the word "location":

Screenshot 2025-01-12 at 10 50 42

Not sure how critical it is, though 🤔 
We can see the full sentence after scrolling a bit to the right.

klaudiagrz (commented):
Leftovers from the previous review

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
-_start:
-		mov rax, 1
-		mov rdx, 2
-		push rax
-		push rdx
+- `push` - pushes the operand into the stack.
+- `pop` - pops the top value from the stack.
 
-		mov rax, [rsp + 8]
+x86_64 processors provide additional instruction that brings affect on the stack. Besides those instruction we also have seen familiar to us:
```
klaudiagrz, line 99:
```suggestion
x86_64 processors provide additional instructions that affect the stack. In addition to these, we’ve also seen instructions that are already familiar to us:
```

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
 
-		;;
-		;; Do something
-		;;
+- `call`
+- `ret`
```
klaudiagrz, line 102:
```suggestion
- `call` - calls the given procedure. It affects the stack by saving the return address before the call.
- `ret` - exits the given procedure. It affects the stack by removing the return address and transferring the execution flow back to it. 
```

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
+
+The first one instruction calls the given procedure. It affects stack by saving the return address on the stack before call. The second instruction is an "exit" from the given procedure. It affects the stack by removing the return address from the stack and transferring the execution flow to it.
```
klaudiagrz, line 104:
```suggestion

```

I would explain these instructions right in the list as in the comment above. If you like this idea, please remove this paragraph.

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
 
-		;;
-		;; Do something
-		;;
+- `call`
+- `ret`
+
+The first one instruction calls the given procedure. It affects stack by saving the return address on the stack before call. The second instruction is an "exit" from the given procedure. It affects the stack by removing the return address from the stack and transferring the execution flow to it. 
+
+In the [previous post](asm_2.md) we got familiar with the with such a concepts as [function prologue and epilogue](https://en.wikipedia.org/wiki/Function_prologue_and_epilogue). These are special instructions that we usually can meet in the beginning and in the end of the function:
```
klaudiagrz, line 106:
```suggestion
In the [previous post](asm_2.md), we became familiar with concepts such as the [function prologue and epilogue](https://en.wikipedia.org/wiki/Function_prologue_and_epilogue). These are special instructions typically found at the beginning and end of a function:
```

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
+        ;;
+
+        ;; Function epilogue
+        mov rsp, rbp
+        pop
 ```
-Here we can see that we put 1 to rax register and 2 to rdx register. After it we push to stack values of these registers. Stack works as LIFO (Last In First Out). So after this stack or our application will have following structure:
 
-![stack diagram](/content/assets/stack-diagram.png)
+These two could be replaced with special instructions: `enter N, 0` and `leave`. The first instruction has two operands:
```
klaudiagrz, line 123:
```suggestion
These two can be replaced with special instructions: `enter N, 0` and `leave`. The `enter` instruction has two operands: 
```

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
+        pop
 ```
-Here we can see that we put 1 to rax register and 2 to rdx register. After it we push to stack values of these registers. Stack works as LIFO (Last In First Out). So after this stack or our application will have following structure:
 
-![stack diagram](/content/assets/stack-diagram.png)
+These two could be replaced with special instructions: `enter N, 0` and `leave`. The first instruction has two operands: 
 
-Then we copy value from stack which has address rsp + 8. It means we get address of top of stack, add 8 to it and copy data by this address to rax. After it rax value will be 1.
+- Number of bytes that needs to be subtracted from the `rsp` register to allocate space on stack.
+- Number of levels of stack frames in nested calls.
```
klaudiagrz, line 126:
```suggestion
- Number of bytes to subtract from the `rsp` register to allocate space on the stack.
- Number of stack frame levels in nested calls.
```

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
 
-## Example
+These both instructions are considered "outdated" but still will work because of backward compatibility.
```
klaudiagrz, line 128:
```suggestion
These instructions are considered "outdated" but still work because of backward compatibility.
```

klaudiagrz: Why are they outdated? By what are they replaced?

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
 
-Let's see one example. We will write simple program, which will get two command line arguments. Will get sum of this arguments and print result.
+The next already familiar to us instruction that affects the stack is the `syscall` instruction. In some aspects it is similar to the `call` instruction with the one of the most significant difference is that the function that is going to be called is located in the kernel space. The return from a system call and the stack clean-up is executed with the help of the `sysret` instruction.
```
klaudiagrz, line 130:
```suggestion
The next familiar instruction that affects the stack is the `syscall` instruction. In some aspects, it is similar to the `call` instruction, with one key difference: the function to be called is located in kernel space. The return from a system call and the stack clean-up are executed using the `sysret` instruction.
```

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
 
-```assembly
-section .data
-		SYS_WRITE equ 1
-		STD_IN    equ 1
-		SYS_EXIT  equ 60
-		EXIT_CODE equ 0
+In the previous post, we mentioned that besides the general purpose registers, the other types of registers exists. One of such type of registers is `rflags`. In basic words it is a register where CPU stores its current state. In the next posts we will know more details about this type of register but for now we must know that an x86_64 process provide the two following command that affect the stack:
```
klaudiagrz, line 132:
```suggestion
In the previous post, we mentioned that there are other types of registers besides the general purpose registers. One such register is `rflags` where the CPU stores its current state. In the next posts, we will learn more about it. For now, we must know that the x86_64 processor provides the following two commands that affect the stack:
```

content/asm_3.md
```diff
@@ -11,243 +14,604 @@ global _start
...
 
-		NEW_LINE   db 0xa
-		WRONG_ARGC db "Must be two command line argument", 0xa
-```
+- `pushf` - pushes the `rflags` register into the stack.
+- `popf` - pops the top value from the stack and stores the value in the `rflags` register.
```
klaudiagrz, line 135:
```suggestion
- `pushf` - pushes the `rflags` register into the stack.
- `popf` - pops the top value from the stack and stores it in the `rflags` register.
```

klaudiagrz (commented):
Review of the "Stack operations" section

content/asm_3.md
0xAX: Will remove

content/asm_3.md
0xAX: Good note, added the reason

content/asm_3.md
```diff
@@ -11,243 +14,603 @@ global _start
...
+    ;; Number of the standard output file descriptor
+	STD_OUT	equ 1
+    ;; Exit code from the program. The 0 status code is a success.
+	EXIT_CODE equ 0
+    ;; ASCII code of the new line symbol ('\n')
+	NEW_LINE db 0xa
+    ;; Error message that is printed in a case of not enough command line arguments
+	WRONG_ARGC_MSG	db "Error: expected two command line argument", 0xa
+    ;; Length of the WRONG_ARGC_MSG message
+    WRONG_ARGC_MSG_LEN equ 42
```
Yimura, line 161: indentation

content/asm_3.md
```diff
@@ -11,243 +14,603 @@ global _start
...
 		jne exit
 		;;
-		;; Do something
+		;; Otherwise, perform another action.
 		;;
 
 incRax:
+        ;; Increment the value of the rax register
 		inc rax
+        ;; Return from the incRax subroutine
```
Yimura, line 32: indentation

Yimura (changes requested):
I came here because I was quite confused by the `div` instruction (where does the quotient go and what happens to the dropped number/remainder of the integer?). 

Basically what I've done is fixed some of the sentence structuring and typo's related to this.
Review comments with `indentation` in them means that you've used mixed indentations across lines. For example the below piece of assembly:
```asm
    ;; This will give us the numeric value of the character.
	sub bl, 48
```
The comment is indented with spaces and the assembly is indented with a HARD tab `\t`.
To the compiler this makes no difference but for the reader it comes off as messy.

0xAX:
@Yimura thank you very much for your comments, I will fix in the near time.

content/asm_3.md
```diff
@@ -11,243 +14,603 @@ global _start
...
+    ;; Return from the str_to_int procedure.
+	ret
+
+;; Convert the sum to string and print it on the screen.
+int_to_str:
+	;; High part of the dividend. The low part is in the rax register.
+	mov rdx, 0
+	;; Set the divisor to 10.
+	mov rbx, 10
+	;; Divide the sum (rax from rax) to 10. Reminder will be stored in the rdx register.
```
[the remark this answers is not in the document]
0xAX: Thank you for suggestion. Applied along with other comments. Thank you one more time.

content/asm_3.md
```diff
@@ -11,243 +14,603 @@ global _start
...
-		NEW_LINE   db 0xa
-		WRONG_ARGC db "Must be two command line argument", 0xa
-```
+- `pushf` - pushes the `rflags` register into the stack.
+- `popf` - pops the top value from the stack and stores it in the `rflags` register.
 
-First of all we define `.data` section with some values. Here we have four constants for linux syscalls, for sys_write, sys_exit and etc... And also we have two strings: First is just new line symbol and second is error message.
+## Example
+
+After we went through some theory it is time to write some code! Let's see another one example that should make us more confident with the assembly programming. This time we will write a simple program, which will get [two command line arguments](https://en.wikipedia.org/wiki/Command-line_interface#Arguments), try to calculate sum of the given values and print the result.
```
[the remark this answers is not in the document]
0xAX: Yes, it is good to add :+1:

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
+	mov r11, rax
+
+	;; Calculate the sum of the arguments. The result will be stored in the r10 register.
+	add r10, r11
+
+	;; Move the sum value to the rax register.
+	mov rax, r10
+	;; Initialize counter by resetting it to 0. It will store the length of the result string.
+	xor rcx, rcx
+	;; Convert the sum from number to string to print the result on the screen.
```
klaudiagrz, line 56:
```suggestion
	;; Convert the sum from a number to a string to print the result on the screen.
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
-	mov	rax, r10
-	;; number counter
-	xor	r12, r12
-	;; convert to string
-	jmp	int_to_str
-
-;;
-;; Print argc error
-;;
+	;; Fetch the number of arguments from the stack and store it in the rcx register
```
klaudiagrz, line 25:
```suggestion
	;; Fetch the number of arguments from the stack and store it in the rcx register.
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
-	add	rax, rbx
-	;; get next number
-	inc	rsi
-	;; again
-	jmp	next
-
-return_str:
+	mov rcx,  10
+__repeat:
+	;; Compare the first element in the given string with the NUL terminator (end of string).
```
klaudiagrz, line 81:
```suggestion
	;; Compare the first element in the given string with the NUL terminator (end of the string).
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
+	;; Subtract the value 48 from the ASCII code of the current character.
+	;; This will give us the numeric value of the character.
+	sub bl, 48
+	;; Multiple our result number by 10 to get the place for the next digit.
+	mul rcx
+	;; Add the next digit to our result number.
+	add rax, rbx
+	;; Move to the next character in the command line argument string.
+	inc rsi
+	;; Repeat until we do not reach the end of the string.
```
klaudiagrz, line 96:
```suggestion
	;; Repeat until we reach the end of the string.
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
-	mul	r12
-	mov	rdx, rax
-	;;;;
-
-	;;;; print sum
-	mov	rax, SYS_WRITE
-	mov	rdi, STD_IN
-	mov	rsi, rsp
-	;; call sys_write
+;; Convert the sum to string and print it on the screen.
```
klaudiagrz, line 102:
```suggestion
;; Convert the sum to a string and print it on the screen.
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
+int_to_str:
+	;; High part of the dividend. The low part is in the rax register.
+	mov rdx, 0
+	;; Set the divisor to 10.
+	mov rbx, 10
+	;; Divide the sum stored in `rax`, resulting quotient will be stored in `rax`,
+	;; and the remainder will be stored in `rdx` register.
```
klaudiagrz, line 109:
```suggestion
	;; Divide the sum stored in `rax. The resulting quotient will be stored in `rax`,
	;; and the remainder will be stored in the `rdx` register.
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
+	div rbx
+	;; Add 48 to the remainder to get a string ASCII representation of the number value.
+	add rdx, 48
+	;; Store the remainder on the stack.
+	push rdx
+	;; Increase the counter.
+	inc rcx
+	;; Compare the rest of the sum with zero.
+	cmp rax, 0x0
+	;; If it is not zero yet, continue to convert it to string.
```
klaudiagrz, line 119:
```suggestion
	;; If it is not zero, continue to convert it to string.
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
+	jne int_to_str
+	;; Otherwise print the result.
```
klaudiagrz, line 121:
```suggestion
	;; Otherwise, print the result.
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
+	jmp printResult
+
+;; Print the result to the standard output.
+printResult:
+	;; Put the number of symbols within the string to the rax register.
```
klaudiagrz, line 126:
```suggestion
	;; Put the number of string characters to the rax register.
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
 section .data
-	SYS_WRITE	equ 1
-	STD_IN		equ 1
-	SYS_EXIT	equ 60
-	EXIT_CODE	equ 0
-
-	NEW_LINE	db 0xa
-	WRONG_ARGC	db "Must be two command line argument", 0xa
-
+	;; Number of `sys_write` system call
```
klaudiagrz, line 3:
```suggestion
	;; Number of the `sys_write` system call
```

stack/stack.asm
```diff
@@ -1,168 +1,162 @@
...
+	SYS_WRITE equ 1
+	;; Number of `sys_exit` system call
```
klaudiagrz, line 5:
```suggestion
	;; Number of the `sys_exit` system call
```

klaudiagrz (commented):
Comments to the code sample

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
+2. Take the first byte of the string and subtract the value `48` from it. Each byte in a string is an [ASCII](https://en.wikipedia.org/wiki/ASCII) symbol with its own code. The symbol 0 has code `48`, the symbol 1 has code `49`, and so on. If we subtract `48` from the ASCII code of the given symbol, we get an integer representation of the current digit from the given string.
+3. As soon as we know the current digit, we multiply our accumulator from step 1 by 10 and add to it the digit that we got in step 2.
+4. Move to the next symbol in the given string and repeat steps 2 and 3 if it is not the end of the string (`\0` symbol).
+
+Returning to the table from the section above, we may see that pointers to the command line arguments are located on the stack right above the number of command line arguments. So if we fetch the first value from the stack after we already fetched the number of arguments, it will be a pointer to the string which is the first command line argument. If we will pop the next value from the stack, it will be the second command line argument passed to the program.
```
klaudiagrz, line 408:
```suggestion
Returning to the table from the section above, we can see that pointers to the command-line arguments are located on the stack right above the number of command-line arguments. So, after we pop the number of arguments (ARGC), the stack pointer will point to the address of the first command-line argument (ARGV[0]). If we pop the next value from the stack, it will point to the second command-line argument (ARGV[1]) passed to the program.
```

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
+
+Now if we will take a look at the `str_to_int` procedure it should be clear without any additional details:
```
klaudiagrz, line 410:
```suggestion
Now the `str_to_int` procedure should be more clear:
```

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
+	inc rsi
+	;; Repeat until we do not reach the end of the string.
+	jmp __repeat
+__return:
+	;; Return from the str_to_int procedure.
+	ret
 ```
 
-At the start of str_to_int, we set up rax to 0 and rcx to 10. Then we go to next label. As you can see in above example (first line before first call of str_to_int) we put argv[1] in rsi from stack. Now we compare first byte of rsi with 0, because every string ends with NULL symbol and if it is we return. If it is not 0 we copy it's value to one byte bl register and substract 48 from it. Why 48? All numbers from 0 to 9 have 48 to 57 codes in asci table. So if we substract from number symbol 48 (for example from 57) we get number. Then we multiply rax on rcx (which has value - 10). After this we increment rsi for getting next byte and loop again. Algorthm is simple. For example if rsi points to '5' '7' '6' '\000' sequence, then will be following steps:
+As soon as we converted both command line arguments to integer numbers, we can calculate their sum:
```
klaudiagrz, line 460:
```suggestion
As soon as we converted both command-line arguments to integer numbers, we can calculate their sum:
```

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
-    rax * 10 --> rax = 0 * 10
-    rax = rax + rbx = 0 + 5
-    Get second byte - 7 and put it to rbx
-    rax * 10 --> rax = 5 * 10 = 50
-    rax = rax + rbx = 50 + 7 = 57
-    and loop it while rsi is not \000
-```
 
-After str_to_int we will have number in rax. Now let's look at int_to_str:
+Since we have our result, we just need to print it. But before printing it we have to convert the numeric result to string. This we will see in the next section.
```
klaudiagrz, line 467:
```suggestion
Now that we have our result, we just need to print it. But before printing it, we have to convert the numeric result back to a string. This we will learn in the next section.
```

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
+
+### Converting integer to string
```
klaudiagrz, line 469:
```suggestion
### Converting an integer to a string
```

klaudiagrz (commented):
Suggestions till "Converting integer to string"

content/assets/asm-3-args-on-stack.svg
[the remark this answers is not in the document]
0xAX: Nice catch

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
-    123 / 10. rax = 12; rdx = 3
-    rdx + 48 = "3"
-    push "3" to stack
-    compare rax with 0 if no go again
-    12 / 10. rax = 1; rdx = 2
-    rdx + 48 = "2"
-    push "2" to stack
-    compare rax with 0, if yes we can finish function execution and we will have "2" "3" ... in stack
-```
+The algorithm of the `int_to_str` sobroutine is pretty simple as well. We divide our number by `10` to get the digit and add the value `48` to the result of the division. Remember about ASCII codes? If yes it should be clear why we are doing it. As soon as we got the symbolic representation of the current digit we push it on the stack. As soon as the given digit is converted we increase our counter of numbers of symbols within the string and check our sum number. If it is zero it means we have the resulted string. If not, we just repeat the all operations.
```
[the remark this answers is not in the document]
klaudiagrz: If yes, it should be clear why we are doing it:
1. As soon as we get the symbolic representation of the current digit, we push it on the stack. 
2. When the given digit is converted, we increase our counter that represents the number of characters within the string. 
3. After that, we check the sum number. If it is zero, we have the resulting string. If not, we repeat all operations.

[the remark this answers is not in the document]
klaudiagrz: ^ consider the alternative formating of text in the form of ordered list

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
 	mov rax, SYS_EXIT
-	exit code
+	;; Set the first argument of `sys_exit` to 0. The 0 status code is a success.
 	mov rdi, EXIT_CODE
+	;; Call the `sys_exit` system call.
 	syscall
 ```
 
-That's All.
+Most of this code should be already well understandable for you as the most significant part of it consists of the initialization of data for the call of the `sys_write` and `sys_exit` exit calls. The most interesting part should be first four lines of code of the `printResult` subroutine. As you may remember the one of the parameters of the `sys_write` system call is a length of the string that we want to print on the screen. We have this number as we maintained a counter of symbols during converting the numeric sum to the string. This counter was stored in the `rcx` register. Our string is located on the stack. We pushed each digit with the `push` operator. But the `push` operator pushes `64` bits (or `8` bytes) while our symbol is only 1 byte. To get the whole length of the string for printing, we should multiple the number of symbols to `8`. This will give us the length of the string that we can use as a third argument of the `sys_write` system call.
```
klaudiagrz, line 554:
```suggestion
Most of this code should already be understandable, as it mainly consists of the data initialization for the `sys_write` and `sys_exit` system calls. The most interesting part is the first four lines of the `printResult` subroutine. As you may remember, one of the parameters of the `sys_write` system call is the length of the string we want to print on the screen. We have this number because we maintained a counter of characters while converting the numeric sum to a string. This counter was stored in the `rcx` register. Our string is located on the stack, where we pushed each digit using the `push` operator. However, the `push` operator pushes `64` bits (or `8` bytes), while our symbol is only 1 byte. To calculate the total length of the string for printing, we should multiply the number of symbols by `8`. This will give us the length of the string that we can use as the third argument of the `sys_write` system call.
```

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
+
+As soon as all parameters of the system calls are ready, we can pass them as arguments to print the sum and print new line after it.
```
klaudiagrz, line 556:
```suggestion
Once all parameters of both system calls are ready, we can pass them as arguments to print the sum followed by a new line.
```

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
+
+Let's build our program with the usual commands:
```
klaudiagrz, line 558:
```suggestion
Now, let's build our program with the usual commands:
```

content/asm_3.md
```diff
@@ -11,243 +14,609 @@ global _start
...
+
+```bash
+$ nasm -f elf64 -o stack.o stack.asm
+$ ld -o stack stack.o
+```
+
+And try to run it:
```
klaudiagrz, line 565:
```suggestion
Then, try to run it:
```

klaudiagrz (commented):
Comments till "Security considerations"

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+}
+
+int main() {
+    foo();
+    printf("Program exited successfully\n");
+    return 0;
+}
+```
+
+If we will try to build this program and run it, we'll see the following error instead of the *Program exited successfully* string:
```
klaudiagrz, line 611:
```suggestion
If we build and run this program, we'll see the following error instead of the `Program exited successfully` string:
```

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+
+```bash
+$ ./test
+Enter text: 123456789
+*** stack smashing detected ***: terminated
+Aborted (core dumped)
+```
+
+The reason for this is that we put on the stack the value which is bigger than our 8 bytes buffer. Happily instead of overwriting of return address or segmentation fault error we have got "stack smashing detected" error. This check is done by the modern compiler to prevent overwriting of critical data. There are other techniques in modern compilers and operating system kernels to mitigate vulnerabilities related to stack, like:
```
klaudiagrz, line 620:
```suggestion
The reason for this error is that we put on the stack a value bigger than our 8-byte buffer. Happily, instead of overwriting the return address or segmentation fault error, we get a `stack smashing detected` error. This check is done by a modern compiler to prevent overwriting critical data. There are also other techniques in modern compilers and operating system kernels to mitigate vulnerabilities related to stack, like:
```

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+
+- [Stack canaries](https://en.wikipedia.org/wiki/Buffer_overflow_protection#Canaries)
+- [ASLR](https://en.wikipedia.org/wiki/Address_space_layout_randomization)
+- [Non-executable stack](https://en.wikipedia.org/wiki/Executable-space_protection)
+- And others...
+
+In any cases, despite all of these techniques may help you to protect your programs from stack related errors, you should be careful, especially with the data that your program receives from outside.
```
klaudiagrz, line 627:
```suggestion
Despite all of these techniques may help you to protect your programs from stack-related errors, you should be careful, especially with the external data that your program receives.
```

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+
+This example might be a little bit artificial as unlikely you are going to use the `gets` function in your code. The [manual page](https://man7.org/linux/man-pages/man3/gets.3.html) of this function says:
+
+> Never use gets().  Because it is impossible to tell without knowing the data in advance how many characters gets() will read, and because gets() will continue to store characters past the end of the buffer, it is extremely dangerous to use.  It has been used to break computer security.  Use fgets() instead.
+
+Moreover this function is deprecated. But despite such artificial example, there can be real danger even if you are not using deprecated functions and use all the options of the compiler that help you to protect your program.
```
klaudiagrz, line 633:
```suggestion
This example might seem a bit artificial as unlikely you are going to use the deprecated `gets` function. However, even with such an unrealistic example, real risks still exist — even if you avoid deprecated functions and use all the compiler’s safety features to protect your program.
```

klaudiagrz: too much "screen time" for `gets`

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
...53). This vulnerability was found in the Linux kernel and led to the [privilege escalation](https://en.wikipedia.org/wiki/Privilege_escalation). When the kernel runs a process, it needs to perform many different operations. One of such operations are to load the program into memory and initialize the stack. The binary itself is loaded and located below the stack memory. Besides that, there is a gap which is 128 megabytes. The loading of a big enough binary led to the situation when certain segments of the binary were loaded and mapped to this gap. The binaries with the big enough data segements (bingger than 128 megabytes) can end up mapped over the stack memory. All of this may lead to the situation when the [ELF .dynamic section](https://en.wikipedia.org/wiki/Executable_and_Linkable_Format) is overwritten by the stack. One may put a path to a shared library that will be loaded by the kernel and the code executed with the higher privileges. If you are interested in more details you can read the [report](https://www.qualys.com/2017/09/26/linux-pie-cve-2017-1000253/cve-2017-1000253.txt) and the [fix](https://github.com/torvalds/linux/commit/a87938b2e246b81b4fb713edb371a9fa3c5c3c86).
```
klaudiagrz, line 635:
```suggestion
The real-world case when wrong memory management led to serious consequences is [CVE-2017-1000253](https://nvd.nist.gov/vuln/detail/CVE-2017-1000253). This vulnerability was found in the Linux kernel and led to the [privilege escalation](https://en.wikipedia.org/wiki/Privilege_escalation). When the kernel runs a process, it needs to perform many different operations, such as loading the program into memory and initializing the stack. After the program is loaded and stack initialized, the program is located below the stack memory, with a 128-megabyte gap between them. However, when a large program is loaded, it can overwrite the stack memory. Under certain conditions, it may lead to privilege escalation. If you are interested in more details, you can read the [report](https://www.qualys.com/2017/09/26/linux-pie-cve-2017-1000253/cve-2017-1000253.txt) and the [fix](https://github.com/torvalds/linux/commit/a87938b2e246b81b4fb713edb371a9fa3c5c3c86).
```

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+
+As we've seen, subtle bugs in stack layout can lead to serious vulnerabilities.
```
klaudiagrz, line 637:
```suggestion
As you can see, subtle bugs in stack layout can lead to serious vulnerabilities.
```

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+
+## Conclusion
+
+We’ve just written our third program using assembly — great job 🎉 In the next post, we’ll continue exploring assembly programming and see more details how to work with strings. If you have any questions or thoughts, feel free to reach out. See you in the next post!
```
klaudiagrz, line 641:
```suggestion
We’ve just written our third program using assembly — great job 🎉 In the next post, we’ll continue exploring assembly programming and see more details on how to work with strings. If you have any questions or thoughts, feel free to reach out. See you in the next post!
```

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
-	    mul rcx
-	    add rax, rbx
-	    inc rsi
-	    jmp next
-
-return_str:
-	    ret
+	;; Set the value of the rax register to 0. It will store the result.
+	xor rax, rax
+	;; base for multiplication
```
klaudiagrz, line 435:
```suggestion
	;; Base for multiplication
```

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+$ ./stack 5 10
+15
+```
+
+Works as expected 🎉🎉🎉
+
+
+## Security considerations
+
+As seen in this and the previous posts, the stack is a crucial concept used to manage function calls in our programs. Understanding how the stack memory is managed is important for writing programs with reusable functions and crucial for writing secure programs. The stack is a common source of security vulnerabilities, especially in low-level code and assembly routines. When you use `call` and `ret` instructions, the processor doesn’t verify if the return address is valid, but it just simply pops the address and jumps on it. One of the most common problems is the [stack overflow](https://en.wikipedia.org/wiki/Stack_overflow).
```
klaudiagrz, line 588:
```suggestion
As seen in this and the previous posts, the stack is a crucial concept used to manage function calls in our programs. Understanding how the stack memory is managed is important for writing programs with reusable functions and crucial for writing secure programs. The stack is a common source of security vulnerabilities, especially in low-level code and assembly routines. When you use `call` and `ret` instructions, the processor doesn’t verify if the return address is valid, but it simply pops the address and jumps on it. One of the most common problems is the [stack overflow](https://en.wikipedia.org/wiki/Stack_overflow).
```

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+The reason for this error is that we put on the stack a value bigger than our 8-byte buffer. Happily, instead of overwriting the return address or segmentation fault error, we get a `stack smashing detected` error. This check is done by a modern compiler to prevent overwriting critical data. There are also other techniques in modern compilers and operating system kernels to mitigate vulnerabilities related to stack, like:
+
+- [Stack canaries](https://en.wikipedia.org/wiki/Buffer_overflow_protection#Canaries)
+- [ASLR](https://en.wikipedia.org/wiki/Address_space_layout_randomization)
+- [Non-executable stack](https://en.wikipedia.org/wiki/Executable-space_protection)
+- And others...
+
+Despite all of these techniques may help you to protect your programs from stack-related errors, you should be careful, especially with the external data that your program receives.
+
+This example might be a little bit artificial as unlikely you are going to use the `gets` function in your code. The [manual page](https://man7.org/linux/man-pages/man3/gets.3.html) of this function says:
```
klaudiagrz, line 629: Which example? I feel like we lost a paragraph somewhere 🤔

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+
+> Never use gets().  Because it is impossible to tell without knowing the data in advance how many characters gets() will read, and because gets() will continue to store characters past the end of the buffer, it is extremely dangerous to use.  It has been used to break computer security.  Use fgets() instead.
+
+This example might seem a bit artificial as unlikely you are going to use the deprecated `gets` function. However, even with such an unrealistic example, real risks still exist — even if you avoid deprecated functions and use all the compiler’s safety features to protect your program.
```
klaudiagrz, line 633: And here we repeat the paragraph above 🤔

content/asm_3.md
```diff
@@ -1,7 +1,10 @@
...
+In the [previous post](asm_2.md), we started to learn the basics of the x86_64 architecture. One of the most crucial concepts we learned was the [stack](https://en.wikipedia.org/wiki/Stack-based_memory_allocation). In this chapter, we will explore more examples of stack usage.
 
-We have 16 general-purpose registers for temporary data storage. They are RAX, RBX, RCX, RDX, RDI, RSI, RBP, RSP and R8-R15. It's too few for serious applications. So we can store data in the stack. Yet another usage of stack is following: When we call a function, return address copied in stack. After end of function execution, address copied in commands counter (RIP) and application continue to executes from next place after function.
+Let's start with a quick remainder: the stack is a special memory region that operates on the LIFO (last-in, first-out) principle. In the x86_64 architecture, we have sixteen general-purpose registers for temporary data storage: `rax`, `rbx`, `rcx`, `rdx`, `rdi`, `rsi`, `rbp`, `rsp`, and from `r8` to `r15`. However, for some applications, this might not be enough. One way to overcome this limitation is by using the stack.
```
klaudiagrz, line 5:
```suggestion
Let's start with a quick reminder: the stack is a special memory region that operates on the LIFO (last-in, first-out) principle. In the x86_64 architecture, we have sixteen general-purpose registers for temporary data storage: `rax`, `rbx`, `rcx`, `rdx`, `rdi`, `rsi`, `rbp`, `rsp`, and from `r8` to `r15`. However, for some applications, this might not be enough. One way to overcome this limitation is by using the stack.
```

content/asm_3.md
```diff
@@ -11,243 +14,628 @@ global _start
...
+	jmp exit
 
-		add rsp, 8
-		pop rsi
-		call str_to_int
+;; Convert the command line argument to the integer number.
+str_to_int:
+	;; Set the value of the rax register to 0. It will store the result.
+	xor rax, rax
+	;; base for multiplication
```
klaudiagrz, line 222:
```suggestion
	;; Base for multiplication
```

content/asm_3.md
```diff
@@ -11,243 +14,626 @@ global _start
...
+The reason for this error is that we put on the stack a value bigger than our 8-byte buffer. Happily, instead of overwriting the return address or segmentation fault error, we get a `stack smashing detected` error. This check is done by a modern compiler to prevent overwriting critical data. There are also other techniques in modern compilers and operating system kernels to mitigate vulnerabilities related to stack, like:
+
+- [Stack canaries](https://en.wikipedia.org/wiki/Buffer_overflow_protection#Canaries)
+- [ASLR](https://en.wikipedia.org/wiki/Address_space_layout_randomization)
+- [Non-executable stack](https://en.wikipedia.org/wiki/Executable-space_protection)
+- And others...
+
+Despite all of these techniques may help you to protect your programs from stack-related errors, you should be careful, especially with the external data that your program receives.
+
+This example might be a little bit artificial as unlikely you are going to use the `gets` function in your code. The [manual page](https://man7.org/linux/man-pages/man3/gets.3.html) of this function says:
```
klaudiagrz, line 629:
```suggestion
The C function example above might be a little bit artificial as unlikely you are going to use the `gets` function in your code. The [manual page](https://man7.org/linux/man-pages/man3/gets.3.html) of this function says:
```

klaudiagrz (approved):
Good job! 🎉

PR #47, opened 2025-06-01
asm-5: Rework of the 5th chapter

0xAX:
**Description**

This PR provides rework of the 5th chapter.


content/asm_5.md
```diff
@@ -1,138 +1,448 @@
...
-cmp rax, 3
-jne .mustBe3args
+;; Store the number of command line arguments in the rax register
+mov rax, argc
+;; Store the pointer to the first command like argument in the rsi register
+mov rsi, argv1
 ```
 
-Multiline macro starts with %macro nasm directive and end with %endmacro. It general form is following:
+If we use the `%define` directive to define a macro, the macro code is evaluated when the macro is used. NASM provides additional directive - `%xdefine` which has the same form as the `%define` directive but it is evaluated immediately at the point of definition.
```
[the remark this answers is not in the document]
0xAX: Yes... expanded would be more correct

content/asm_5.md
```diff
@@ -1,138 +1,448 @@
...
+Now let's try to go through the macros definitions and try to understand how they work. The macro definition starts from the `%macro` directive and definition of the name of the macro and number of input parameters. Our `PRINT` macro will expect two arguments:
+
+- The reference to the string that is going to be printed
+- The length of this string
+
+The `EXIT` macro expects only one input argument - exit code. 
+
+There is an additional syntax - `n-*` which says that a macro accepts at least `n` arguments.
+
+In the macros body we just initialize the registers according to the [ABI](https://refspecs.linuxbase.org/elf/x86_64-abi-0.99.pdf) to call the [system calls](https://en.wikipedia.org/wiki/System_call). The only one difference that you may note is that instead of direct values, we use the input parameters in the macro. The input parameters starts from the `%` symbol followed by the number of the parameter.
```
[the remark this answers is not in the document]
0xAX: Yes, will add

content/asm_5.md
```diff
@@ -1,138 +1,448 @@
...
+
+For example:
+
+```assembly
+%assign increment_i_var i + 1
+```
+
+### %defstr directive
+
+The `%defstr` directive exists to define a macro without parameters that will expand to a quoted string.
```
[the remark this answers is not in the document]
0xAX: without is correct here

content/asm_5.md
```diff
@@ -1,138 +1,448 @@
...
+
+### User defined error directives
+
+NASM provides the following three directives to specify user-defined errors or warnings:
+
+- `%error msg`
+- `%fatal msg`
+- `%warning msg`
+
+In a case of the first two the execution of the program will be interrupted if one of the will be met.
```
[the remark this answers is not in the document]
0xAX: Yes, I will add information

content/asm_5.md
```diff
@@ -1,138 +1,448 @@
...
+%macro EXIT 1
+        ;; Specify the number of the system call (60 is `sys_exit`).
+        mov rax, 60
+        ;; Set the first argument of `sys_exit` to the first argument of the macro.
+        mov rdi, %1
+        ;; Call the `sys_exit` system call.
+        syscall
+%endmacro
+
+;; Define person structure
```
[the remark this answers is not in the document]
0xAX: Yes, I like it.

content/asm_5.md
```diff
@@ -1,138 +1,449 @@
...
+        PRINT msg_2, 9
+        ;; Print new line message with the length 1.
+        PRINT newline, 1
+        ;; Exit from the program. The 0 status code is success.
+        EXIT 0
+```
+
+Now let's go through the macro definitions and see how they work. A macro definition starts from the `%macro` directive, followed by the macro’s name and the number of input parameters it expects. For example, our `PRINT` macro expects two arguments:
+
+- The reference to the string that is going to be printed
```
klaudiagrz, line 161:
```suggestion
- The reference to the string that will be printed
```

klaudiagrz: ... or `that we want to print`

content/asm_5.md
```diff
@@ -1,138 +1,449 @@
...
+- The length of this string
+
+The `EXIT` macro expects only one input argument - exit code. 
+
+There is also an additional syntax (`n-*`) that indicates a macro accepts at least `n` arguments.
+
+In the macro's body, we initialize the registers according to the [ABI](https://refspecs.linuxbase.org/elf/x86_64-abi-0.99.pdf) to make [system calls](https://en.wikipedia.org/wiki/System_call). The only difference you may notice is that instead of using direct values, we use input parameters. These parameters start with the `%` symbol, followed by the number of the parameter (`%1`, `%2`, etc).
+
+Another syntax ability in the multi-line NASM macros is the possibility to define labels inside a macro. In this case, the label name should be prefixed with `%%`. For example:
```
klaudiagrz, line 170: Should we mention why anyone would need to define labels? Or is it rather obvious?

content/asm_5.md
```diff
@@ -1,138 +1,449 @@
...
+
+```bash
+~$ nasm -f elf64 -o test.o test.asm && ld -o test test.o && ./test
+Message 1
+```
+
+However, if we pass `DEBUG` to the program or define it, we will see that both messages are printed:
+
+```bash
+~$ nasm -DDEBUG -f elf64 -o test.o test.asm && ld -o test test.o && ./test
```
klaudiagrz, line 222:
```suggestion
~$ nasm -DEBUG -f elf64 -o test.o test.asm && ld -o test test.o && ./test
```

content/asm_5.md
```diff
@@ -1,138 +1,449 @@
...
+
+### User-defined error directives
+
+NASM provides the following directives to specify user-defined errors or warnings:
+
+- `%error msg`
+- `%fatal msg`
+- `%warning msg`
+
+For the first two directives, the program will print the message given as an argument of the directive and the execution will interrupted. In a case of the `%warning` directive, only the message will be printed.
```
klaudiagrz, line 275:
```suggestion
For the first two directives, the program prints the message given as an argument, and the program's execution is interrupted. In the case of the `%warning` directive, only the message is printed.
```

content/asm_5.md
```diff
@@ -1,138 +1,449 @@
...
+        %rotate 1
+        %%f(%1)
+    %endrep
+%endmacro
 ```
 
-### %include
+As you may see, this macro contains comments that help to understand from the beginning what it should do. Let's try to understand how this macro does its job.
+
+First of all, the definition of the macro is a little bit new to us. The `2-*` notation means that the macro accepts at least 2 input parameters. The macro will use the first argument as an instruction and the rest of the arguments to which this instruction will be applied.
```
klaudiagrz, line 411:
```suggestion
First of all, the definition of the macro is a little bit new to us. The `2-*` notation means that the macro accepts at least 2 input parameters. The macro uses the first argument as an instruction and applies it to the remaining arguments.
```

content/asm_5.md
```diff
@@ -1,138 +1,449 @@
...
+
+The first line in the macro body defines the local macro named `%%f`. This macro is expanded to the first argument of the `REPX` macro. Because the `xdefine` directive is used instead of `%define`, it captures the value of the first argument of the `REPX` macro as is. For example, `%%f` will be bound to `{psrlw x, 8}`, if the macro is used like this:
+
+```assembly
+REPX {psrlw x, 8}, m0, m1, m2, m3
+```
+
+The macro name is prefixed with `%%` to define a uniquely scoped macro with the name which scoped only to this invocation of the `REPX` macro to not clash with other macros named `f` elsewhere.
```
klaudiagrz, line 419:
```suggestion
The macro name is prefixed with `%%` to define a uniquely scoped macro, with the name limited to this invocation of the `REPX` macro. This prevents clashes with other macros named `f` elsewhere.
```

content/asm_5.md
```diff
@@ -1,138 +1,449 @@
...
+
+The loop in this macro consists of two lines of code:
+
+- The first line contains the `%rotate` directive, which rotates the arguments by 1 position at each iteration of the loop. It moves the second argument to the first position, the third argument to the second position, and so on.
+- The second line has the call of the `%%f` macro. Remember that this macro is bound to the first original instruction of the `REPX` macro. By passing the `%1` to the `%%f` macro it expands to the original first argument of the `REPX` macro with the `x` replaced with the first argument of the same macro but after we rotated them. So at the first iteration of the loop it will be expanded to the first argument of the `REPX` macro with `x` replaced with the second argument of the same macro. At the second iteration of the loop it will be expanded to the first argument of the `REPX` macro with `x` replaced with the third argument of the same macro. This will be repeated for the all arguments of the `REPX` macro.
```
klaudiagrz, line 426:
```suggestion
- The second line calls the `%%f` macro, which was previously defined to represent the first argument of the `REPX` macro. By passing `%1` to `%%f`, the macro expands that first argument, replacing the placeholder `x` with each of the subsequent arguments to `REPX`, one at a time. On each iteration of the loop, `x` is substituted with the next argument — so the first iteration uses the second argument, the second iteration uses the third, and so on. This is repeated until all arguments are processed.
```

content/asm_5.md
```diff
@@ -1,138 +1,449 @@
...
+It will be expanded to:
+
+```assembly
+pmulhuw m0, m7
+pmulhuw m1, m7
+pmulhuw m2, m7
+pmulhuw m3, m7
+```
+
+We have just tried to understand real-world assembly code!
```
klaudiagrz, line 443:
```suggestion
We’ve just explored a real-world assembly code! 🎉 🎉 🎉 
```

klaudiagrz (commented):
Comments after re-reading

content/asm_5.md
```diff
@@ -1,138 +1,449 @@
...
+
+The loop in this macro consists of two lines of code:
+
+- The first line contains the `%rotate` directive, which rotates the arguments by 1 position at each iteration of the loop. It moves the second argument to the first position, the third argument to the second position, and so on.
+- The second line has the invocation of the %%f macro, which was previously defined to expand into the first argument of the REPX macro. It also replaces x in the first argument of REPX with the parameter given to the %%f macro. By passing %1 to %%f, the macro expands to the original first argument of the REPX macro, replacing the placeholder x with each of the subsequent arguments of the REPX. On each iteration of the loop, x is substituted with the next argument of the REPX macro starting from the second because of the previous rotate definition  — so the first iteration uses the second argument, the second iteration uses the third, and so on. This is repeated until all arguments are processed.
```
klaudiagrz, line 426:
```suggestion
- The second line invokes the `%%f` macro, which was previously defined to expand into the first argument of the REPX macro. It also replaces `x` in the first argument of REPX with the parameter given to the `%%f` macro. By passing `%1` to `%%f`, the macro expands to the original first argument of the REPX macro, replacing the placeholder `x` with each of the subsequent arguments of REPX. On each iteration of the loop, `x` is substituted with the next argument of the REPX macro starting from the second because of the previous rotate definition — so the first iteration uses the second argument, the second iteration uses the third, and so on. This is repeated until all arguments are processed.
```

klaudiagrz (approved):
Yay! Please apply the comment with formatting and merge. Well done! 🎉

klaudiagrz (approved):

PR #51, opened 2025-06-06
Fix the example in the 4th part

0xAX:
**Description**

This PR provides the fix for the example in the 4th part. The problem was that it describes as the `rdi` register as a storage for the length of the reversed string while it is actually has an end address of the output buffer.

This commit fixes the example to use the length of the string.

### DIFF
--- a/content/asm_4.md
+++ b/content/asm_4.md
@@ -192,7 +192,7 @@ _start:
         ;; Store the address of the input string in the rsi register.
         mov rsi, INPUT
         ;; Store the address of the output buffer in the rdi register.
-        mov  rdi, OUTPUT
+        mov rdi, OUTPUT
         ;; Call the reverseStringAndPrint procedure.
         call reverseStringAndPrint
 ```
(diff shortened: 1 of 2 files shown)


content/asm_4.md
```diff
@@ -210,6 +210,8 @@ After that, we need to call the `reverseStringAndPrint` procedure to calculate t
 reverseStringAndPrint:
         ;; Compare the first element in the given string with the NUL terminator (end of the string).
         cmp byte [rsi], 0
+        ;; Preserve the length of the reversed string in the rdx register. We will use this value during printing the string.
```
klaudiagrz, line 213:
```suggestion
        ;; Preserve the length of the reversed string in the rdx register. We will use this value when printing the string.
```

strings/reverse.asm
```diff
@@ -33,14 +33,16 @@ _start:
...
-        mov  rdi, OUTPUT
+        mov rdi, OUTPUT
         ;; Call the reverseStringAndPrint procedure.
         call reverseStringAndPrint
 
 ;; Calculate the length of the input string and prepare to reverse it.
 reverseStringAndPrint:
         ;; Compare the first element in the given string with the NUL terminator (end of the string).
         cmp byte [rsi], 0
+        ;; Preserve the length of the reversed string in the rdx register. We will use this value during printing the string.
```
klaudiagrz, line 44:
```suggestion
        ;; Preserve the length of the reversed string in the rdx register. We will use this value when printing the string.
```

klaudiagrz (approved):
Feel free to merge after applying the comments :)

PR #56, opened 2025-06-22
Rework 6th chapter

0xAX:
**Description**

This PR provides full rework of the 6th chapter.


0xAX:
The link-checker is failed and it is ok to fail as it has 1 link that points to the "future" assembly file that is going to be added only in this PR.

content/asm_6.md
```diff
@@ -1,181 +1,535 @@
...
-```assembly
-.data
-    //
-    // initialized data definition
-    //
-.text
-    .global _start
+| Sign | Exponent | Mantissa                |
+|------|----------|-------------------------|
+|     0| 10000001 | 01100100110011001100110 |
```
klaudiagrz, line 92: I do not understand why this table is different from the one above :(

content/asm_6.md
```diff
@@ -1,181 +1,535 @@
...
 
-_start:
-    //
-    // main routine
-    //
-```
+Now we know how computer represents a floating point number with a single-precision 🎉
```
klaudiagrz, line 94:
```suggestion
Now we know how a computer represents a floating point number with a single-precision 🎉 
```

content/asm_6.md
```diff
@@ -1,181 +1,535 @@
...
-_start:
-    //
-    // main routine
-    //
-```
+Now we know how computer represents a floating point number with a single-precision 🎉 
+
+### Double and double-extended precision formats
+
+As you may guess, the representation of the floating-point numbers using the double-precision and double-extended precision formats is similar to the single-precision format and the difference is only accuracy.
```
klaudiagrz, line 98:
```suggestion
As you may guess, the representation of floating-point numbers using the double-precision and double-extended precision formats is similar to the single-precision format, and the only difference is accuracy.
```

content/asm_6.md
```diff
@@ -1,181 +1,535 @@
...
+
+The double-precision format is `64-bits` of memory with the following structure:
+
+- sign - 1 bit
+- exponent - 11 bit
+- mantissa - 52 bit
```
klaudiagrz, line 104:
```suggestion
- Sign - 1 bit
- Exponent - 11 bit
- Mantissa - 52 bit
```

content/asm_6.md
```diff
@@ -1,181 +1,535 @@
...
+
+The extended-double precision format is `80-bits` of memory with the following structure:
+
+- sign - 1 bit
+- exponent - 15 bit
+- mantissa - 112 bit
```
klaudiagrz, line 110:
```suggestion
- Sign - 1 bit
- Exponent - 15 bit
- Mantissa - 112 bit
```

content/asm_6.md
```diff
@@ -1,181 +1,535 @@
...
-    // main routine
-    //
-```
+Now we know how computer represents a floating point number with a single-precision 🎉 
+
+### Double and double-extended precision formats
+
+As you may guess, the representation of the floating-point numbers using the double-precision and double-extended precision formats is similar to the single-precision format and the difference is only accuracy.
+
+The double-precision format is `64-bits` of memory with the following structure:
```
klaudiagrz, line 100:
```suggestion
The double-precision format uses `64-bits` of memory and has the following structure:
```

content/asm_6.md
```diff
@@ -1,181 +1,535 @@
...
+
+- sign - 1 bit
+- exponent - 11 bit
+- mantissa - 52 bit
+
+The extended-double precision format is `80-bits` of memory with the following structure:
```
klaudiagrz, line 106:
```suggestion
The extended-double precision format uses `80-bits` of memory and has the following structure:
```

content/asm_6.md
klaudiagrz: How much is it for single-precision format?

content/asm_6.md
klaudiagrz: Are these values correct? Shouldn't they sum up to give 80? (that was the case for double-precision format).

content/asm_6.md
```diff
@@ -1,181 +1,535 @@
...
+
+- sign - 1 bit
+- exponent - 15 bit
+- mantissa - 112 bit
+
+The `bias` for the double-precision format is `1023`. The `bias` for the extended-double precision is `16383`.
```
klaudiagrz, line 112:
```suggestion
The `bias` for the double-precision format is `1023`. The `bias` for the extended-double precision format is `16383`.
```

klaudiagrz (commented):
Review pt.1

content/asm_6.md
```diff
@@ -1,181 +1,535 @@
...
-.LFE0:
-	.size	main, .-main
-	.ident	"GCC: (Ubuntu 4.9.1-16ubuntu6) 4.9.1"
-	.section	.note.GNU-stack,"",@progbits
+>>> 0.2 + 0.4
+0.6000000000000001
 ```
 
-Looks different then nasm Hello world, let's look on some differences.
+After we converted both integral and fractional parts of our number to binary representation, we can go to the next step. We need to convert to [scientific notation](https://en.wikipedia.org/wiki/Scientific_notation). The first part we need to [shift right](https://en.wikipedia.org/wiki/Logical_shift) the integral part of our number so that only `1` will remain. In a case of our number `101` - we need to shift right the number 2 times. As the result our initial number `101.101` started to be represented as `1.01101 * 2^2`. After this we need to add the number of shifts we done to the special number called - `bias`. For single-precision floating-point numbers it is equal to `127`. So we will get - `2 + 127 = 129` or `0b10000001`. This will be our `exponent`. The `mantissa` is just the fractional part of the number that we got after shifting it. We just put all the number of the fractional parts to the 23 `mantissa` bits.
```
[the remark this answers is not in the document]
klaudiagrz: Parts that are not understandable:
- Where this one comes from: `2^2`
- Where this one comes from and it means: `0b10000001`

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+- `movss` - Move single-precision floating-point value between the `XMM` registers or between an `XMM` register and memory,
+- `movsd` - Move double-precision floating-point value between the `XMM` registers or between an `XMM` register and memory.
+- `movhlps` - Move two packed single-precision floating-point values from the high quadword of an `XMM` register to the low quadword of another `XMM` register.
+- `movlhps` - Move two packed single-precision floating-point values from the low quadword of an `XMM` register to the high quadword of another `XMM` register.
+- And others.
+
+### Floating-point arithmetic instructions
 
-You can note two little differences here:
+The floating-point arithmetic instructions perform arithmetic operations such as add, subtract, multiplication, and division on single or double precision floating-point values. The following floating-point arithmetic instructions exist:
```
klaudiagrz, line 161:
```suggestion
The floating-point arithmetic instructions perform arithmetic operations, such as addition, subtraction, multiplication, and division, on single or double-precision floating-point values. The following floating-point arithmetic instructions exist:
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
 
-* Section definition starts with . symbol
-* Main routine defines with .globl instead global as we do it in nasm
+- `addss` - Add single-precision floating-point values.
+- `addsd` - Add dobule-precision floating point values.
```
klaudiagrz, line 164:
```suggestion
- `addsd` - Add double-precision floating point values.
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+- `subss` - Subtract single-precision floating-point values.
+- `subsd` - Subtract double-precision floating-point values.
+- `mulss` - Multiply single-precision floating-point values.
+- `mulsd` - Multiply double-precision floating-point values.
+- `divss` - Divide single-precision floating-point values. 
+- `divsd` - Divide double-precision floating-point values.
```
klaudiagrz, line 170: ... or 3rd person singular? (`-s`)

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
 
-Also gas uses another directives for data defintion:
+All of these instructions expects two operands to execute the given operation of addition, subtraction, multiplication, or division. After the operation will be executed, the result will be stored in the first operand.
```
klaudiagrz, line 172:
```suggestion
All of these instructions expect two operands to execute the given operation of addition, subtraction, multiplication, or division. After the operation is executed, the result is stored in the first operand.
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+
+### Floating-Point control instructions
```
klaudiagrz, line 174:
```suggestion
### Floating-point control instructions
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+
+As we already know, the main goal of this type of instructions is to manage the control flow of our programs. For the integer comparison we have seen the `cmp` instruction. But this instruction will not work with floating-point values. Although, the result of the comparison is controlled by the same `rflags` registers that we have seen in the [4th part](./asm_4.md#Control transfer instructions).
```
klaudiagrz, line 176:
```suggestion
As we already know, the main goal of these instructions is to manage the control flow of our programs. For integer comparisons, we used the `cmp` instruction. But this instruction does not work with floating-point values, even though the result of the comparison is controlled by the same `rflags` registers that we saw in the [Data manipulation](./asm_4.md#Control transfer instructions) chapter.
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+
+The general instruction to compare floating-point numbers is `cmpss`.
```
klaudiagrz, line 178: I'd join it with the previous paragraph, as it is closely related. Also... isn't any information missing? Seems like this section ended too suddenly.

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+
+The calling conventions for floating-point numbers is different. If a function parameter is a floating-point number - it will be passed in one of `XMM` registers - the first argument will be passed in the `xmm0` register, the second argument will be passed in `xmm1` register and so on up till `xmm7`. The rest of arguments after eights argument are passed on the [stack](./asm_3.md). The value of the functions that return a floating-point value is stored in the `xmm0` register.
+
+### Data transfer instructions
+
+As we know from the [4th part](./asm_4.md) - data transfer instructions are used to move data between different locations. `x86_64` CPU provides special set of instructions for transfer of floating point data. If you are going to use the `x87 FPU` unit, one of the most common data transfer instructions are:
+
+- `fld` - Load floating point value.
+- `fst` - Store floating point value.
```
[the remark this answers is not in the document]
0xAX: it is instruction, so yes

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+
+These instructions are used by the `x87 FPU` unit to store and load the floating-point numbers at/from the stack.
+
+To work with `XMM` registers, the following data transfer instructions are supported by an `x86_64` CPU:
+
+- `movss` - Move single-precision floating-point value between the `XMM` registers or between an `XMM` register and memory,
+- `movsd` - Move double-precision floating-point value between the `XMM` registers or between an `XMM` register and memory.
+- `movhlps` - Move two packed single-precision floating-point values from the high quadword of an `XMM` register to the low quadword of another `XMM` register.
+- `movlhps` - Move two packed single-precision floating-point values from the low quadword of an `XMM` register to the high quadword of another `XMM` register.
+- And others.
```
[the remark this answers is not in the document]
0xAX: Can we add something here to not have impressions that there are just four of them?

float/README.md
```diff
@@ -1,7 +1,11 @@
-This is simple application which shows how to work with floating point numbers.
+# Dot product
 
-Build it with: `make`
+This is a simple program that read two double vectors and calculates [dot product](https://en.wikipedia.org/wiki/Dot_product) of their values.
```
klaudiagrz, line 3:
```suggestion
This is a simple program that reads two double vectors and calculates the [dot product](https://en.wikipedia.org/wiki/Dot_product) of their values.
```

float/dot_product.asm
```diff
@@ -0,0 +1,304 @@
...
+        ;; Number of the standard input file descriptor.
+        STD_IN equ 0
+        ;; Number of the standard output file descriptor.
+        STD_OUT equ 1
+        ;; Exit code from the program. The 0 status code is a success.
+        EXIT_CODE equ 0
+
+        ;; Maximum number of elements in a vector
+        MAX_ELEMS equ 100
+        ;; Size of buffer that we will use to read vectors
```
klaudiagrz, line 18:
```suggestion
        ;; Size of the buffer that we will use to read vectors
```

float/dot_product.asm
```diff
@@ -0,0 +1,304 @@
...
+;; Definition of the .text section
+section .text
+        ;; Reference to the C stdlib functions that we will use
+        extern strtod, printf
+        ;; Reference to the entry point of our program.
+        global _start
+
+;; Entry point of the program.
+_start:
+        ;; Read the first input string with floating-point values
```
klaudiagrz, line 67:
```suggestion
        ;; Proceed to the _read_first_float_vector section
```

float/dot_product.asm
```diff
@@ -0,0 +1,304 @@
...
+        ;; Initialize the rsi register with the pointer which will point to the place where
+        ;; the strtod(3) will finish its work.
+        mov rsi, end_buffer_1
+        ;; Call the strtod(3) to conver floating-point value from the input buffer to double.
+        call strtod
+
+        ;; Preserve the pointer to the next floating-point value from the input buffer
+        ;; in the rax register.
+        mov rax, [end_buffer_1]
+        ;; Check is it the end of the input string.
```
klaudiagrz, line 121:
```suggestion
        ;; Check whether it is the end of the input string.
```

float/dot_product.asm
```diff
@@ -0,0 +1,304 @@
...
+        ;; Initialize the rsi register with the pointer which will point to the place where
+        ;; the strtod(3) will finish its work.
+        mov rsi, end_buffer_2
+        ;; Call the strtod(3)
+        call strtod
+
+        ;; Preserve the pointer to the next floating-point value from the input buffer
+        ;; in the rax register.
+        mov rax, [end_buffer_2]
+        ;; Check is it the end of the input string.
```
klaudiagrz, line 198:
```suggestion
        ;; Check whether it is the end of the input string.
```

float/dot_product.asm
```diff
@@ -0,0 +1,304 @@
...
+        cmp rax, rdi
+        ;; Calculate the dot product after we have both vectors.
+        je _calculate_dot_product
+        
+        ;; Store the reference to the beginning of the buffer where we will store
+        ;; our double values to the rdx register.
+        mov rdx, vector_2
+        ;; Store the number of floating-point values we already have read in the rcx register.
+        mov rcx, r15
+        ;; Multiple the number of floating-point values by 8.
```
klaudiagrz, line 208:
```suggestion
        ;; Multiply the number of floating-point values by 8.
```

klaudiagrz (commented):
Comments to code example

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+        ;; Specify the system call number (1 is `sys_write`).
+        mov rax, SYS_WRITE
+        ;; Set the first argument of `sys_write` to 1 (`stdout`).
+        mov rdi, STD_OUT
+        ;; Set the second argument of `sys_write` to the reference of the prompt string to print.
+        mov rsi, FIRST_INPUT_MSG
+        ;; Call the `sys_write` system call.
+        syscall
+
+        ;; Set the length of string we want to read from the standard input.
```
klaudiagrz, line 311:
```suggestion
        ;; Set the length of the string we want to read from the standard input.
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+        ;; read the data for the vector.
+        mov rsi, buffer_1
+        ;; Call the `sys_read` system call.
+        syscall
+
+        ;; Save the number of bytes we have read from the standard input in the rcx register.
+        mov rcx, rax
+        ;; Set the pointer to the beginning of the buffer with the input data to the rdx register.
+        mov rdx, buffer_1
+        ;; Move pointer within the buffer to the end of input.
```
klaudiagrz, line 327:
```suggestion
        ;; Move the pointer within the buffer to the end of input.
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+        add rdx, rcx
+        ;; Fill the last byte of the input with 0.
+        mov byte [rdx], 0
 ```
 
-Also you can not here that registers starts with % symbol. If you're using direct operands, need to use `$` symbol:
+The code starts from the already familiar to us call of the `sys_write` system call. We use it to write a prompt string `Input first vector: ` to the terminal. After this line is printed, we execute the `sys_read` system call to read the user input. If we will take a look at the definition of this system call, we will see that it expects to get three parameters:
```
klaudiagrz, line 333:
```suggestion
As usual, the code starts from the `sys_write` system call. We use it to write a prompt string `Input first vector: ` to the terminal. After this line is printed, we execute the `sys_read` system call to read the user input. If we take a look at the definition of this system call, we will see that it expects three arguments:
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
-movb $10, %rax
+```C
+ssize_t read(int fd, void buf[.count], size_t count);
 ```
 
-### Size of operands and operation syntax
+The parameters are:
 
-Sometimes when we need to get part of memory, for example first byte of 64 register, we used following syntax:
+- A file descriptor which identifies file that we want to read the data from
```
klaudiagrz, line 341:
```suggestion
- A file descriptor that identifies the file from which we want to read the data
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+- The buffer where the data that we have read will be stored
+- Number of bytes that we want to read
 
-```assembly
-mov ax, word [rsi]
-```
+You may see that we specify all of these parameters before calling of the `sys_read` in our code according to the **calling conventions** specified in the [System V Application Binary Interface](https://refspecs.linuxbase.org/elf/x86_64-abi-0.99.pdf). If you already forgot the details, you can read this document or the [second part](./asm_2.md) of this notes.
```
klaudiagrz, line 345:
```suggestion
As you can see, we specify all these parameters before calling `sys_read` according to the **calling conventions** specified in the [System V Application Binary Interface](https://refspecs.linuxbase.org/elf/x86_64-abi-0.99.pdf). If this sounds unfamiliar, read the linked document or go back to the [second chapter](./asm_2.md) of this assembler course.
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
 
-There is another way for such operations in gas. We don't define size in operands but in instruction:
+After this code will be executed, our buffer specified by the `buffer_1` name, will contain the user input. Now we have the line that a user passed to our program, the next goal is to convert each value represented by the string from the input to the floating-point values and store them in a separate buffer. But before we can do it, we need to do one more action. The user input will contain the `newline` symbol in the end. We don't need it in our input as we can't convert it to a floating point number. To get rid of this symbol - we just replace it with the `0` byte. To do that we need just know the length of the user-input. The good news that we already know it. If you will take a look at the documentation of the `sys_read` system call, you may see:
```
klaudiagrz, line 347:
```suggestion
After executing the code, the buffer specified by the `buffer_1` name will contain the user input. The next goal is to convert each value represented by the string from the input to floating-point values and store them in a separate buffer. But before, we need to do one more thing. 

The user input contains the `newline` symbol at the end. We don't need it in our input, as we can't convert it to a floating-point number. To get rid of this symbol, we replace it with the `0` byte. To do that, we need to know the length of the user input. The good news is that we already know it. Take a look at the documentation of the `sys_read` system call:
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+
+> On success, the number of bytes read is returned
+
+As you may remember, the return value of a system call is stored in the `rax` register. To write the zero byte into the buffer right after the user input, we just need to take the pointer to the beginning of this buffer, add offset to it equal to the length of the user-input and write `0` byte to this address. All of these you may see in the last four lines of code above.
```
klaudiagrz, line 351:
```suggestion
As you may remember, the return value of a system call is stored in the `rax` register. To write the zero byte into the buffer right after the user input, we just need to take the pointer to the beginning of this buffer, add an offset to it (which is equal to the length of the user input), and add the `0` byte to this address. All of these you can see in the last four lines of the code above.
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+
+Now we have buffer with the string values and this means that we can start to convert them to the floating point numbers. We will see how to do it in the next section!
```
klaudiagrz, line 353:
```suggestion
Now we have a buffer with the string values. This means that we can start converting them to floating-point numbers. We will see how to do it in the next section!
```

content/asm_6.md
```diff
@@ -1,181 +1,544 @@
...
+Input first vector: 2.5 3.17
+Input second vector: 4.22 100.1
+Dot product = 327.867000
 ```
+
+Works as expected 🎉🎉🎉
+
+## Conclusion
+
+In this chapter, we have seen how to work with a floating-point data and got familiar with more assembly instructions. Of course, this small post didn't cover all the details related to this big topic.
```
klaudiagrz, line 542:
```suggestion
In this chapter, we have seen how to work with floating-point data and got familiar with more assembly instructions. Of course, this post didn't cover all the details related to this topic.
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+        ;; Set address of the second vector to the rdi register.
+        mov rsi, vector_2
+        ;; Set the number of values within the vectors to the rdx register.
+        mov rdx, r14
+        ;; Calculate the dot product of the two vectors.
+        call _dot_product
+```
 
-You can note that we used () brackets in previous example instead [] in nasm example. To dereference values in parentheses are used GAS: (%rax), for example:
+Before the calculation of the dot product of two vectors we must be sure that both vectors have the same number of components. The number of components of the first vector we stored in the `r14` register and the number of components of the second vector we stored in the `r15` register. Let's' compare them and if they are not equal, let's print error and exit. The error printing and the exit from the program should be already familiar to you:
```
klaudiagrz, line 455:
```suggestion
Before calculating the dot product of two vectors, we must be sure that both vectors have the same number of components. The number of components of the first vector is stored in the `r14` register, and the number of components of the second vector is stored in the `r15` register. Let’s check if these values are equal; if not, we will print an error message and exit the program. The error-printing and program exit process should already be familiar to you:
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+    ;; Specify the number of the system call (60 is `sys_exit`).
+    mov rax, SYS_EXIT
+    ;; Set the first argument of `sys_exit` to 0. The 0 status code is success.
+    mov rdi, EXIT_CODE
+    ;; Call the `sys_exit` system call.
+    syscall
 ```
 
-### Jumps
+If our data is good, we can proceed to calculation. To do that we store the address of both our vectors in the `rdi` and `rsi` registers, and put the number of components within the vectors into the `rdx` register. The `_dot_product` function does the main job. Let's take a look at the code of this function:
```
klaudiagrz, line 483:
```suggestion
When both vectors have the same number of components, we can proceed to the calculation. To do that, we store the addresses of both vectors in the `rdi` and `rsi` registers, and put the number of components within the vectors into the `rdx` register. The `_dot_product` function does the main job. Let's take a look at the code of this function:
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+```assembly
+;; Calculate the dot product of the two vectors.
+_dot_product:
+        ;; Reset the value of the rax register to 0.
+        xor rax, rax
+        ;; Reset the value of the xmm1 register to 0.
+        pxor xmm1, xmm1
+        ;; Current rdx contains the number of floating-point values within the vectors.
+        ;; Multiply it by 8 to get the number of bytes occupied by these values.
+        sal rdx, 3
```
klaudiagrz, line 494: Leave a comment what `sal` instruction does. Why `3`?

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+;; Calculate the the dot product in the loop.
+_loop:
+        ;; Move the floating-point value from the first vector to the xmm0 register.
+        movsd xmm0, [rdi + rax]
+        ;; Multiply the floating-point from the second vector to the value from the first vector
```
klaudiagrz, line 499:
```suggestion
        ;; Multiply the floating-point value from the second vector by the value from the first vector
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+        ;; and store the result in the xmm0 register.
+        mulsd xmm0, [rsi + rax]
+        ;; Move to the next floating-point values in the vector buffers.
+        add rax, 8
+        ;; Add the result of multiplication of floating-point values from the vectors in the
+        ;; xmm1 register.
```
klaudiagrz, line 505:
```suggestion
        ;; Add the result of multiplying the floating-point values from the two vectors into the xmm1 register
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+        addsd xmm1, xmm0
+        ;; Check if we went through all the floating-point values in the vector buffers.
+        cmp rax, rdx
+        ;; If not yet - repeat the loop.
```
klaudiagrz, line 509:
```suggestion
        ;; If not, repeat the loop.
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+        jne _loop
+        ;; Move the result to the xmm0 register.
+        movapd xmm0, xmm1
+        ;; Return from the _dot_product back to the `_calculate_dot_product`.
```
klaudiagrz, line 513:
```suggestion
        ;; Return from the `_dot_product` back to the `_calculate_dot_product`.
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+        ret
+```
+
+In the beginning of the function we prepare the `rax` and `xmm1` registers by resetting them to zero. The `rax` register will contain the offset within the vector buffers and the `xmm1` will be accumulator which will accumulates result of our program.
```
klaudiagrz, line 517:
```suggestion
At the beginning of the function, we prepare the `rax` and `xmm1` registers by resetting them to zero. The `rax` register will contain the offset within the vector buffers, and the `xmm1` register will accumulate the result of our program.
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+
+All the calculation of the dot products of our two vectors happens within the loop identified by the `_loop` label. In the first instruction after this label, we store the current value from the first vector in the register `xmm0`. After this we multiple this value by the current value from the second vector. Remember that both vectors are pointed by the `rdi` and `rsi` register and the `rax` register stores the offset within this buffers that pointer to the current value we need to process. As we did the first multiplication, we increase the value of the `rax` to eight bytes to move to the next values within the vector buffers. In the same time we update our accumulator with the intermediate result of the dot product.
```
klaudiagrz, line 519:
```suggestion
The dot product calculation takes place inside the loop labeled `_loop`. In the first instruction of this label, we put the current value from the first vector into the `xmm0` register. Next, we multiply this value by the current value from the second vector. The two vectors are pointed by the `rdi` and `rsi` registers, and the `rax` register stores the offset within these buffers that points to the current value we need to process. After performing the multiplication, we increase the value of the `rax` to eight bytes to move to the next values within the vector buffers. At the same time, we update our accumulator with the intermediate result of the dot product.
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
 
-GNU assembler supports following operators for far functions call and jumps:
+At this point, the `xmm1` register will have the result of multiplication of the fist components of our vectors. At the next step we check that - did we reach the end of vectors and if not we repeat the loop for the second components, third, and so on.
```
klaudiagrz, line 521:
```suggestion
At this point, the `xmm1` register stores the multiplication result of the first components of our vectors. Now we must check if we reached the end of vectors; if not, we repeat the loop for the second component, third, and so on.
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+
```
klaudiagrz, line 522: I would remove this line break and join these paragraphs into one

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+As soon as we reached the end of the vectors, we store the result in the `xmm0` register and return from the `_dot_product` function.
```
klaudiagrz, line 523:
```suggestion
As soon as we reach the end of the vectors, we put the result into the `xmm0` register and return from the `_dot_product` function.
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+
+Our result is ready 🎉 🎉 🎉 The last remaining thing to do - is to print the result. We will do using the [printf](https://man7.org/linux/man-pages/man3/printf.3.html) function to simplify our program:
```
klaudiagrz, line 525:
```suggestion
Our result is ready 🎉 🎉 🎉 The last thing to do is to print it. We will do it using the [printf](https://man7.org/linux/man-pages/man3/printf.3.html) function to simplify our program:
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
 
 ```assembly
-lcall $section, $offset
+        ;; Specify reference to the format string for the printf(3) in the rdi register.
```
klaudiagrz, line 528:
```suggestion
        ;; Specify a reference to the format string for the printf(3) in the rdi register.
```

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+        mov rdi, PRINTF_FORMAT
+        ;; Number of arguments of the floating-point registers passed as arguments
+        ;; to printf(3). We specify - `1` because we need to pass only `xmm0` with
```
klaudiagrz, line 531:
```suggestion
        ;; to printf(3). We specify `1` because we need to pass only `xmm0` with
```

klaudiagrz: It still doesn't tell me why 1 :(

content/asm_6.md
```diff
@@ -1,181 +1,563 @@
...
+Now, let's build our program with the usual commands:
 
-### Comments
+```bash
+$ nasm -g -f elf64 -o dot_product.o dot_product.asm
+$ ld -dynamic-linker /lib64/ld-linux-x86-64.so.2 -lc dot_product.o -o dot_product
+```
 
-GNU assembler supports 3 types of comments:
+Then, try to run it:
```
klaudiagrz, line 548:
```suggestion
Then, let's run it:
```

content/asm_6.md
0xAX: added

content/asm_6.md
0xAX: Done

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
+To get the binary representation, we just write down all the integral parts that we got during the calculations. Integral parts for `0.625` are `1`, `0`, and `1`, which gives us `0b0.101` in binary. As a result, a binary representation of our floating point number `5.625` is `0b101.101`.
+
+As you practice converting decimal floating-point numbers to binary, you'll soon notice an interesting pattern: not all fractional parts can be represented in binary. For example, let's take a look at the binary representation of the fractional part of the number `5.575`:
+
+| Multiplication | Result | Integral part | Fractional part |
+|----------------|--------|---------------|-----------------|
+|      0.575 * 2 |   1.15 |             1 |            0.15 |
+|       0.15 * 2 |   0.30 |             0 |            0.30 |
+|       0.30 * 2 |   0.60 |             0 |            0.60 |
+|       0.60.* 2 |   1.20 |             1 |            0.20 |
```
klaudiagrz, line 61:
```suggestion
|       0.60 * 2 |   1.20 |             1 |            0.20 |
```

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
+- `subss` - Subtracts a single-precision floating-point values.
+- `subsd` - Subtracts a double-precision floating-point values.
+- `mulss` - Multiplies a single-precision floating-point values.
+- `mulsd` - Multiplies a double-precision floating-point values.
+- `divss` - Divides a single-precision floating-point values.
+- `divsd` - Divides a double-precision floating-point values.
+
+All of these instructions expect two operands to execute the given operation of addition, subtraction, multiplication, or division. After the operation is executed, the result is stored in the first operand.
+
+### Floating-Point control instructions
```
klaudiagrz, line 173:
```suggestion
### Floating-point control instructions
```

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
+The floating-point arithmetic instructions perform arithmetic operations, such as addition, subtraction, multiplication, and division, on single or double-precision floating-point values. The following floating-point arithmetic instructions exist:
+
+- `addss` - Adds a single-precision floating-point values.
+- `addsd` - Adds a double-precision floating point values.
+- `subss` - Subtracts a single-precision floating-point values.
+- `subsd` - Subtracts a double-precision floating-point values.
+- `mulss` - Multiplies a single-precision floating-point values.
+- `mulsd` - Multiplies a double-precision floating-point values.
+- `divss` - Divides a single-precision floating-point values.
+- `divsd` - Divides a double-precision floating-point values.
```
klaudiagrz, line 169:
```suggestion
- `addss` - Adds single-precision floating-point values.
- `addsd` - Adds double-precision floating point values.
- `subss` - Subtracts single-precision floating-point values.
- `subsd` - Subtracts double-precision floating-point values.
- `mulss` - Multiplies single-precision floating-point values.
- `mulsd` - Multiplies double-precision floating-point values.
- `divss` - Divides single-precision floating-point values.
- `divsd` - Divides double-precision floating-point values.
```

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
+
+All of these instructions expect two operands to execute the given operation of addition, subtraction, multiplication, or division. After the operation is executed, the result is stored in the first operand.
+
+### Floating-Point control instructions
+
+As we already know, the main goal of these instructions is to manage the control flow of our programs. For integer comparisons, we used the `cmp` instruction. But this instruction does not work with floating-point values, even though the result of the comparison is controlled by the same `rflags` registers that we saw in the [Data manipulation](./asm_4.md#control-transfer-instructions) chapter. The general instruction to compare floating-point numbers is `cmpss`. Instead of setting the value of the flag based on the result comparison, the instruction stores the result of the comparison in the destination register.
```
klaudiagrz, line 175:
```suggestion
As we already know, the main goal of these instructions is to manage the control flow of our programs. For integer comparisons, we used the `cmp` instruction. However, this instruction does not work with floating-point values, even though the result of the comparison is controlled by the same `rflags` registers that we saw in the [Data manipulation](./asm_4.md#control-transfer-instructions) chapter. The general instruction to compare floating-point numbers is `cmpss`. Instead of setting the value of the flag based on the result comparison, the instruction stores the result of the comparison in the destination register.
```

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
+
+For example:
 
 ```assembly
-mov destination, source
+;; Compare xmm0 < xmm1 and store the result in the xmm0
```
klaudiagrz, line 180:
```suggestion
;; Compare if xmm0 < xmm1 and store the result in xmm0
```

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
+cmpss xmm0, xmm1, 0x1
 ```
 
-With GNU assembler we have back order i.e.:
+The third operand of the instruction, identifies the operator:
```
klaudiagrz, line 184:
```suggestion
The third operand of the instruction identifies the operator:
```

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
+|                 2 | <=                                                                     |
+|                 3 | Check that one of operands is [Nan](https://en.wikipedia.org/wiki/NaN) |
+|                 4 | !=                                                                     |
+|                 5 | >=                                                                     |
+|                 6 | >                                                                      |
+|                 7 | Checks that both operands are [Nan](https://en.wikipedia.org/wiki/NaN) |
+
+## Example
+
+After we got familiar with the basics of the new topic, it's time to write some code. This time, we will try to write a program which reads the user input, builds two [vectors](https://en.wikipedia.org/wiki/Vector_(mathematics_and_physics)) of floating-point numbers based on it, and calculates the [dot product](https://en.wikipedia.org/wiki/Dot_product) of these two vectors. This operation is widely used in machine learning. It will be interesting to try to implement it using the assembly programming language.
```
klaudiagrz, line 199:
```suggestion
After we got familiar with the basics of the new topic, it's time to write some code. This time, we will try to write a program which reads the user input, builds two [vectors](https://en.wikipedia.org/wiki/Vector_(mathematics_and_physics)) of floating-point numbers based on it, and calculates the [dot product](https://en.wikipedia.org/wiki/Dot_product) of these two vectors. This operation is widely used in machine learning. It will be interesting to implement it using the assembly programming language.
```

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
-mov rax, rcx
+- The prompt messages used when asking a user to input data for vectors
+- The error message to print
+- Parameters of the buffer used to store the user input
 
-//
-// gas syntax
-//
-mov %rcx, %rax
+After defining the data that we may initialize, we need to define uninitialized variables:
```
klaudiagrz, line 256:
```suggestion
After defining the data that we can initialize, we need to define uninitialized variables:
```

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
+In the `.bss` section, we define:
+
+- Two buffers for the vectors
+- Two buffers for the user input
+- Two pointers to the current position within the buffers with the user input
+
+The last parameter here is the most interesting. To simplify our job, we will use the functions from the [C standard library](https://en.wikipedia.org/wiki/C_standard_library). One of such functions is [strtod](https://man7.org/linux/man-pages/man3/strtod.3.html), which converts the given string to a floating point number. It takes two parameters:
+
+- Input string which should be converted to a floating point number
+- The pointer that will point to the first character after the parsed number within the string specified by the parameter above
```
klaudiagrz, line 288:
```suggestion
- The pointer to the first character after the parsed number within the string specified by the parameter above
```

content/asm_6.md
```diff
@@ -1,181 +1,561 @@
...
+        ;; If not, repeat the loop.
+        jne _loop
+        ;; Move the result to the xmm0 register.
+        movapd xmm0, xmm1
+        ;; Return from the `_dot_product` back to the `_calculate_dot_product`.
+        ret
 ```
 
-Far jump - a jump to an instruction located in a different segment than the current code segment but at the same privilege level, sometimes referred to as an intersegment jump.
+At the beginning of the function, we prepare the `rax` and `xmm1` registers by resetting them to zero. The `rax` register will contain the offset within the vector buffers, and the `xmm1` register will accumulate the result of our program. The number of floating-point values within the vectors is stored in the `rdx` register. Since we will move the pointer within the buffers, we need to know how many bytes occupied by these values. To do this we multiple the value of the `rdx` register by `8`. To do this we use the `sal` instruction which executes [arithmetic left shift](https://en.wikipedia.org/wiki/Arithmetic_shift) on the given operand.
```
klaudiagrz, line 518:
```suggestion
At the beginning of the function, we prepare the `rax` and `xmm1` registers by resetting them to zero. The `rax` register will contain the offset within the vector buffers, and the `xmm1` register will accumulate the result of our program. The number of floating-point values within the vectors is stored in the `rdx` register. Since we will move the pointer within the buffers, we need to know how many bytes are occupied by these values. To do this, we multiply the value of the `rdx` register by `8` using the `sal` instruction, which executes [arithmetic left shift](https://en.wikipedia.org/wiki/Arithmetic_shift) on the given operand.
```

klaudiagrz (approved):
Yay, well done 🎉 🎉 🎉

PR #62, opened 2025-09-03
Reworth the 7th chapter

0xAX:
Besides the 7th chapter this PR provides minor fixes for other places.


content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
-```
+- `rdi` - used to pass the first argument to a function.
+- `rsi` - used to pass the second argument to a function.
+- `rdx` - used to pass the third argument to a function.
+- `rcx` - used to pass the fourth argument to a function.
+- `r8` - used to pass the fifth argument to a function.
+- `r9` - used to pass the sixth argument to a function.
 
-## Inline assembly
+If there is more parameters than six, the rest of it will be passed on the [stack](./asm_2.md).
```
klaudiagrz, line 103:
```suggestion
If there are more than six parameters, the rest of them are passed on the [stack](./asm_2.md).
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
 ```
-The typical use of Extended asm statements is to manipulate input values to produce output values. However, your asm statements may also produce side effects. If so, you may need to use the volatile qualifier to disable certain optimizations
+./casm
+hello, world!
 ```
 
-Each operand is described by constraint string followed by C expression in parentheses. There are a number of constraints:
+## Call assembly routines from C
+
+After we have seen how to call C functions from assembly, it’s time to flip the perspective. Now let's try to figure out how assembly code can be exposed as a function and then called from a higher-level language. This is quite important to understand as I mentioned in the beginning of this chapter - in practice assembly is rarely a standalone program. Much more often its pieces live inside a larger project written on a higher level programming language. By writing small assembly routines and calling them like normal functions, it let us achieve performance optimizations beyond what the compiler might provide to us or use the capabilities that are simply not accessible from a higher-level language.
```
klaudiagrz, line 121:
```suggestion
After we saw how to call C functions from assembly, it’s time to flip the perspective. Now, let's try to figure out how to expose assembly code as a function and call it from a higher-level language. As already mentioned, assembly is rarely a standalone program. Much more often, its pieces live inside a larger project written in a higher-level programming language. Writing small assembly routines and calling them like normal functions lets us:
- Achieve performance optimizations beyond what the compiler provides 
- Use the capabilities that are simply not accessible from a higher-level language
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+
+There are two main ways to use assembly code from within C:
+
+1. Calling an assembly function directly
+2. Embedding inline assembly inside C source code
```
klaudiagrz, line 126:
```suggestion
- Calling an assembly function directly
- Embedding inline assembly inside C source code
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
 
-	return 0;
+    printf("The argument length is - %d\n", my_strlen(argv[1]));
+
+    return 0;
 }
 ```
 
-Here we can see the same 2 variables as in previous example and inline assembly definition. First of all we put 1 to rax and rdi registers (write system call number, and stdout) as we did it in our plain assembly hello world. Next we do similar operation with rsi and rdi registers but first operands starts with % symbol instead $. It means str is the output operand referred by %1 and len second output operand referred by %2, so we put values of str and len to rsi and rdi with %n notation, where n is number of output operand. Also there is %% prefixed to the register name.
+This is straightforward C code which I think not needed to be explained in details. It checks that exactly one command-line argument was passed, then calls the `my_strlen` function to determine its length. The result is printed to standard output. As I said - pretty straightforward. The single part which interests us is the implementation of the `my_strlen` function. This function is written in assembly, but not in C.
```
klaudiagrz, line 150:
```suggestion
This is straightforward C code that checks if exactly one command-line argument was passed, and then calls the `my_strlen` function to determine its length. The result is printed to standard output. The single part that interests us here is the implementation of the `my_strlen` function. This function is written in assembly, not in C.
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+
+Let's take a look at the implementation of this function:
 
+```assembly
+;; Definition of the text section
+section .text
+        ;; Reference to the entry point of our program
+        global my_strlen
+
+;; Function that returns the length of the string passed via the first argument
```
klaudiagrz, line 160:
```suggestion
;; Function that returns the length of the string passed in the first argument
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+my_strlen:
+        ;; Reset the register value to zero. It will be returned from the
+        ;; function with the result.
```
klaudiagrz, line 163:
```suggestion
        ;; Reset the register value to zero. The value will be returned from the
        ;; function as the result.
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+        xor rax, rax
+.loop:
+        ;; Compare the first element in the given string with the `NUL` terminator (end of the string).
+        cmp byte [rdi + rax], 0
+        ;; If we reached the `NUL` terminator exit from the function.
```
klaudiagrz, line 168:
```suggestion
        ;; If we reached the `NUL` terminator, exit from the function.
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+        je .done
+        ;; Increase the counter which stores the length of the string.
+        inc rax
+        ;; Repeat the operations above while we didn't reach the end of the string.
```
klaudiagrz, line 172:
```suggestion
        ;; Repeat the operations above until we reach the end of the string.
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+        jmp .loop
+.done:
+        ;; Exit from the function and return the result in the `rax` register.
+        ret
 ```
-    This helps GCC to distinguish between the operands and registers. operands have a single % as prefix
+
+The way function's parameters are passed to our assembly function follows the same calling convention which we have seen in the section above. The first argument, which our case is a pointer to the string, passed in the `rdi` register. At the start of the function we clear the value of the `rax` register, since this will serve as our counter. Each time we encounter a character in the string, the loop increments `rax` using it as an accumulator for the string’s length. The loop continues character by character until it reaches the `NUL` terminator, which signals the end of the string. At that point, the value stored in rax is returned to the caller as the final result.
```
klaudiagrz, line 179:
```suggestion
This function's parameters are passed to our assembly function by following the same calling convention as in the section above. The first argument, which in our case is a pointer to the string, is passed in the `rdi` register. At the start of the function, we clear the value of the `rax` register, since it will serve as our counter. Each time we encounter a character in the string, the loop increments `rax`, using it as an accumulator for the string’s length. The loop continues character by character until it reaches the `NUL` terminator, which signals the end of the string. At that point, the value stored in `rax` is returned to the caller as the final result.
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
-build:
-	gcc casm.c -o casm
+./casm hello
+The argument length is - 5
 ```
 
-## Call C from assembly
+## Use inline assembly in C code
+
+In the section above, we already have mentioned that calling an assembly function from a C code is not a single way to use assembly code from within a C program. The second method is to use `inline assembly`. This method allows us to write an assembly code directly in our C code. This can be useful for performance-critical operations, or when you need to access processor instructions that the compiler does not expose through standard C.
```
klaudiagrz, line 197:
```suggestion
In the section above, we already mentioned that calling an assembly function from C code is not the only way to use assembly code from within a C program. The second method is to use inline assembly. This method allows us to write assembly code directly in our C code. This can be useful for performance-critical operations, or when you need to access processor instructions that the compiler does not expose through standard C.
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+
+In [GCC](https://gcc.gnu.org/) (and in other compilers as well), inline assembly uses a special syntax. In its general form it looks like this:
```
klaudiagrz, line 199:
```suggestion
In [GCC](https://gcc.gnu.org/) (and in other compilers as well), inline assembly uses a special syntax:
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+
+```C
+asm [volatile] ("assembly code"
+                : output operands
+                : input operands
+                : clobbers);
+```
+
+The `asm` keyword introduces the inline assembly block. Adding the `volatile` qualifier tells the compiler not to optimize away or reorder this code, which is important if the code has side effects that the compiler cannot see. As the [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html) explains:
```
klaudiagrz, line 208:
```suggestion
The `asm` keyword introduces the inline assembly block. Adding the `volatile` qualifier tells the compiler not to optimize or reorder this code, which is important if the code has side effects that the compiler cannot see. As the [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html) explains:
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+
+> The typical use of Extended asm statements is to manipulate input values to produce output values. However, your asm statements may also produce side effects. If so, you may need to use the volatile qualifier to disable certain optimizations.
+
+After the definition of the assembly code, we can specify input and output operands which describe how C variables should be mapped to registers or memory. Each operand consists of a constraint string followed by the C expression in parentheses. The constraint tells the compiler what kind of location can hold the operand, and ensures that values are moved in and out properly.
```
klaudiagrz, line 212:
```suggestion
After defining the assembly code, we can specify input and output operands that describe how C variables should be mapped to registers or memory. Each operand consists of a constraint string followed by the C expression in parentheses. The constraint tells the compiler what kind of location can hold the operand, and ensures that values are moved in and out properly.
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
 
-And the last method is to call C function from assembly code. For example we have following simple C code with one function which just prints Hello world:
+For example:
+
+- `r` - tells compiler to use a [general-purpose register](./asm_2.md).
+- `g` - tells compiler to use any register, memory, or immediate integer operand.
+- `f` - tells compiler to use a [floating point register](./asm_6.md).
+- `m` - forces compiler to use a memory location.
```
klaudiagrz, line 219:
```suggestion
- `r` - tells the compiler to use a [general-purpose register](./asm_2.md).
- `g` - tells the compiler to use any register, memory, or immediate integer operand.
- `f` - tells the compiler to use a [floating-point register](./asm_6.md).
- `m` - forces the compiler to use a memory location.
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+
+All the supported constraint strings you can find in the [official documentation](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Output-Operands).
```
klaudiagrz, line 221:
```suggestion
You can find all the supported constraint strings in the [official documentation](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Output-Operands).
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+
+Let's try to rewrite our `hello world` program using the inline assembly.
```
klaudiagrz, line 223:
```suggestion
Let's try to rewrite our `hello world` program using the inline assembly:
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
 
-int print() {
-	printf("Hello World\n");
+	__asm__("movq $1, %%rax \n\t" // rax = 1 - Specify the number of the system call (1 is `sys_write`).
+		"movq $1, %%rdi \n\t"     // rdi = 1 - Set the first argument of `sys_write` to 1 (`stdout`).
+		"movq %1, %%rsi \n\t"     // rsi = str - Set the second argument of `sys_write` to the reference of the `str` variable.
+		"movq %2, %%rdx \n\t"     // rdx = len(str) - Set the third argument of `sys_write` to the length of the `str` variable's value.
+		"syscall"                 // Call the `sys_write` system call.
+		: "=g"(ret)               // Return the result in the `ret` variable.
+		: "g"(str), "g" (len));   // Put `str` and `len` variables in any general operand (memory, register or immediate if possible)
```
klaudiagrz, line 240:
```suggestion
		: "g"(str), "g" (len));   // Put `str` and `len` variables in any general operand (memory, register, or immediate, if possible)
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
+
+        printf("Bytes written: %d\n", ret);
 	return 0;
 }
 ```
 
-Now we can define this function as extern in our assembly code and call it with call instruction as we do it much times in previous posts:
+> [!NOTE]
+> In the example above we have used [GNU as](https://sourceware.org/binutils/docs/as.html) assembly which has a slightly different syntax from the [NASM](https://nasm.us/) assembly. The main differences is that order of operands for the `movq` instruction is changed. We move the value of the left operand to right.
```
klaudiagrz, line 248:
```suggestion
> In the example above, we used [GNU `as` assembly](https://sourceware.org/binutils/docs/as.html), which has a slightly different syntax from the [NASM](https://nasm.us/) assembly. The main difference is that the order of operands for the `movq` instruction is reversed, and we move the value of the left operand to the right.
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
-global _start
+We can build the code above using these instructions:
 
-extern print
+```bash
+gcc casm.c -o casm
+```
 
-section .text
+After building we can run it:
```
klaudiagrz, line 256:
```suggestion
After building, we can run it:
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
-Build it with:
+## Conclusion
 
-```
-build:
-	gcc  -c casm.c -o c.o
-	nasm -f elf64 casm.asm -o casm.o
-	ld   -dynamic-linker /lib64/ld-linux-x86-64.so.2 -lc casm.o c.o -o casm
-```
+This was the final chapter of our journey into assembly programming for Linux x86_64 🎉. I hope you enjoyed it and learned something new. Reaching this point is a big achievement — so my congratulations!
```
klaudiagrz, line 266:
```suggestion
This was the final chapter of our journey into assembly programming for Linux x86_64 🎉. I hope you enjoyed it and learned something new. Reaching this point is a big achievement, so congratulations!
```

content/asm_7.md
```diff
@@ -1,153 +1,268 @@
...
 
-and now we can run our third hello world.
+Thank you for taking this journey with me. Keep experimenting, keep breaking and fixing things, and what is the most importantly — have a fun during coding 🚀
```
klaudiagrz, line 268:
```suggestion
Thank you for taking this journey with me. Keep experimenting, keep breaking and fixing things, and what is most important — have fun during coding! 🚀
```

casm/README.md
```diff
@@ -0,0 +1,15 @@
+# Assembly interaction with high-level programming languages
+
+These are three simple programs which provides examples of interaction of assembly and C programming languages:
+
+- [casm1](./casm1/) - call C function from asm
+- [casm2](./casm2/) - gcc inline assembly
+- [casm3](./casm3/) - call asm function from C
```
klaudiagrz, line 7:
```suggestion
- [casm1](./casm1/) - call the C function from asm
- [casm2](./casm2/) - GCC inline assembly
- [casm3](./casm3/) - call the asm function from C
```

content/asm_7.md
```diff
@@ -1,153 +1,269 @@
...
+        printf("Bytes written: %d\n", ret);
+        return 0;
+}
 ```
 
-Build it with:
+> [!NOTE]
+> In the example above, we used [GNU `as` assembly](https://sourceware.org/binutils/docs/as.html), which has a slightly different syntax from the [NASM](https://nasm.us/) assembly. The main difference is that the order of operands for the `movq` instruction is reversed, and we move the value of the left operand to the right.
+
+We can build the code above using these instructions:
```
klaudiagrz, line 251:
```suggestion
We can build the code above using these commands:
```

klaudiagrz (approved):
Yay! 🎉

Issue #64, opened 2025-09-19
dot_product.asm runtime problem


frankchieng1982:
### Existing issues

- [x] I have searched the existing issues.

### Affected code example

https://github.com/0xAX/asm/blob/master/content/asm_6.md

### Issue description

when i debug with gdb run,segmentation fault issue appears.![Image](https://github.com/user-attachments/assets/782d5208-eaab-4cbe-913d-ed29182f5349) then info registers

![Image](https://github.com/user-attachments/assets/d742eae3-7cdb-4219-946d-8e82aff5604e)

![Image](https://github.com/user-attachments/assets/29fa3724-a82c-4c53-aa26-9b8ff8876ec5)

### Actual vs expected result

register memory issue of dot product

0xAX:
hello @frankchieng1982. I can not reproduce the issue:

```
~$ gdb ./dot_product 
The target architecture is set to "i386:x86-64".
Reading symbols from ./dot_product...
(gdb) run
Starting program: /home/alex/dot_product 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Input the first vector: 2.0 3.0
Input the second vector: 
```

Are you using the code 1:1 from https://github.com/0xAX/asm/blob/master/float/dot_product.asm and build instructions from the makefile near to this file or you changed something?

0xAX:
could you please setup breakpoint

frankchieng1982:
```
(gdb) run
Starting program: /home/user/develop/assembly/asm/dot_product 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Input the first vector: 2.0 3.0

Program received signal SIGSEGV, Segmentation fault.
0x00000000004010b8 in _parse_first_float_vector ()
(gdb) disassemble _parse_first_float_vector
Dump of assembler code for function _parse_first_float_vector:
   0x0000000000401088 <+0>:     movabs $0x4056e8,%rsi
   0x0000000000401092 <+10>:    call   0x401010 <strtod@plt>
   0x0000000000401097 <+15>:    mov    0x4056e8,%rax
   0x000000000040109f <+23>:    cmp    %rdi,%rax
   0x00000000004010a2 <+26>:    je     0x4010c4 <_read_second_float_vector>
   0x00000000004010a4 <+28>:    movabs $0x4030a8,%rdx
[...]

PR #77, opened 2026-03-07
asm-6: align stack to 16-byte boundary before C function calls

0xAX:
**Description**

The System V AMD64 ABI requires the stack to be 16-byte aligned before calling any C library function. Without proper alignment, C functions like strtod() and printf() can segfault on systems that strictly enforce ABI requirements (e.g., AMD EPYC processors).

This fix adds a single instruction at _start to force stack alignment. This ensures RSP % 16 == 0 before any call instruction, which after pushing the 8-byte return address results in RSP % 16 == 8 inside the called function - exactly what the ABI specifies.

**Related issues**

Related issue - #64

### DIFF
--- a/content/asm_6.md
+++ b/content/asm_6.md
@@ -303,11 +303,13 @@ section .text
 
 ;; Entry point of the program.
 _start:
+        ;; Align stack to 16-byte boundary at start since we are going to call C functions.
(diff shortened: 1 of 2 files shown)


klaudiagrz (approved):
One small comment

PR #86, opened 2026-05-16
Fix output message.

awatne:
I just fixed a typo.

### DIFF
--- a/content/asm_2.md
+++ b/content/asm_2.md
@@ -482,7 +482,7 @@ After building the program, we can run it with:
 
 ```bash
 ~$ ./program
-Sum is correct
+The sum is correct!
 ```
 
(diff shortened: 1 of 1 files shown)


0xAX:
Hello @awatne, thank you for PR but it needs also to adjust code and not only output. Do you want to do it? I can do otherwise, thank you!

awatne:
Hello. The code already produces this output.

        ;; The message to print if the sum is correct
        msg db "The sum is correct!", 10

It's just that the example shows a different output.

        ~$ ./program
        Sum is correct

The PR changes the second string to be the same as the first.

0xAX:
ah you are right... Sorry, checked wrong first time. Thank you one more time for your PR. Merging.

PR #88, opened 2026-05-19
Fix spelling errors. Change the output message to be the same in every instance.

awatne:
Hello.  
I went through the asm_3 post and found a couple spelling errors. I also noticed that the output message wasn't the same in every instance, so I changed that too. This is how it appears the first time:

        WRONG_ARGC_MSG  db "Error: expected two command-line arguments", 10
        ;; Length of the WRONG_ARGC_MSG message
        WRONG_ARGC_MSG_LEN equ 43

I also saw that you use both 10 and 0xa for newline, but i didn't change that. I thought it was informative.

### DIFF
--- a/content/asm_3.md
+++ b/content/asm_3.md
@@ -251,11 +251,11 @@ int_to_str:
         ;; Set the divisor to 10.
         mov rbx, 10
         ;; Divide the sum stored in `rax`, resulting quotient will be stored in `rax`,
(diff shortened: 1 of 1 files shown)


klaudiagrz (approved):

klaudiagrz:
Hi @awatne ! 👋🏻 

Thank you for the contribution, nice findings!

Issue #21, opened 2024-06-06
Add a script to generate pdf and epub


0xAX:

rcalixte:
Hi, I've started working on this using some prior art from other projects. I have working PDF generation if you think it would be interesting. A few small patches were needed for the markdown files, including converting the SVG files to PNG files so that they would render well. I left the original SVGs in tact. Other than some emojis not rendering (there's also workarounds here if desired too), I think the content itself is fit for consumption.

[introduction-to-assembly.pdf](https://github.com/user-attachments/files/29535762/introduction-to-assembly.pdf)

Let me know if a PR for this change would be welcome or if there are any changes that should be made prior to that or if it's way out of scope and not fit to PR. Either way, thanks for this great resource!

0xAX:
Hello @rcalixte, 

> I have working PDF generation if you think it would be interesting

Yes sure! The example you attached looks really great. With SVG/PNG images yes, understandable, we have similar issues in [linux-insides](https://github.com/0xAX/linux-insides) with PDF generation. Of course any contribution is welcome, please feel free and let me know if you need any help from my side.

Thank you.

rcalixte:
> With SVG/PNG images yes, understandable, we have similar issues in [linux-insides](https://github.com/0xAX/linux-insides) with PDF generation.

I debated adding the same structure there as well. If we can land things here, should we apply the same there too? There are a few files at the top-level but maybe we can structure things differently if desired.



> Of course any contribution is welcome, please feel free and let me know if you need any help from my side.

I'll work on cleaning things up for a PR. Thank you!

0xAX:
I am still not sure which way is the best to handle images. The longest way probably to convert/redraw and maintain  two set of images and I was ok and in that time and now, just out higher priority to make the content not outdated first.

But as I said that is probably the longest way. For now I am ok with any to have something initial and we can see where to go from there, so I don't have a strong opinion like something ust be handled "this" way but not "that"

rcalixte:
Pandoc technically supports an in-place SVG to PNG conversion but it uses a system dependency to accomplish this. I wanted to avoid that since the CI builds happen in a Docker image but we could try to implement it there as an alternative strategy. Let me get the PR open and you can review it for design/maintainability.

rcalixte:
> we have similar issues in [linux-insides](https://github.com/0xAX/linux-insides) with PDF generation

[...]

PR #90, opened 2026-07-06
Add CI for PDF generation

rcalixte:
**Description**

Add CI to automatically generate PDFs on pushes to master with a manual trigger for creation as well.

A CI job sample run is [here](https://github.com/rcalixte/asm/actions/runs/28828617899/job/85497361448) the forked repository contains a release asset with the PDF as well.

Changes proposed in this pull request:

Content modifications are to support PDF generation:

* SVGs were converted to PNGs
* One newline inserted

**Related issues**

* #21


0xAX:
@rcalixte thank you very much! I will take a look tomorrow morning

0xAX:
@rcalixte I tested it and it looks really great! Thank you a lot. I have just two little suggestions that I'll describe in the next comments.

.github/workflows/gen-pdf.yaml
```diff
@@ -0,0 +1,33 @@
...
+      image: pandoc/latex:3
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: docker://pandoc/latex:3
+        timeout-minutes: 10
+        with:
+          entrypoint: './pandoc-docker.sh'
+
+      - id: short_sha
```
0xAX: I would not generate it at each possible commit/PR. I think the best way will be to build PDF by tags (I will push one right after we will merge the PR) but also leave sha-based builds but only for manual builds (based on workflow_dispatch).

Something like this (I did not test it although by myself yet):

```yaml
---
name: Generate PDF using Pandoc

on:
  push:
    tags:
      - 'v*'
  workflow_dispatch:

jobs:
  create_pdf:
    runs-on: ubuntu-latest

    permissions:
      contents: write

    container:
      image: pandoc/latex:3

    steps:
      - uses: actions/checkout@v7

      - uses: docker://pandoc/latex:3
        timeout-minutes: 10
        with:
          entrypoint: './pandoc-docker.sh'

      - id: release_meta
        shell: bash
        run: |
          short_sha="${GITHUB_SHA::7}"

          if [[ "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]]; then
            echo "tag_name=dev-${short_sha}" >> "$GITHUB_OUTPUT"
            echo "prerelease=true" >> "$GITHUB_OUTPUT"
            echo "make_latest=false" >> "$GITHUB_OUTPUT"
          else
            echo "tag_name=${GITHUB_REF_NAME}" >> "$GITHUB_OUTPUT"
            echo "prerelease=false" >> "$GITHUB_OUTPUT"
            echo "make_latest=true" >> "$GITHUB_OUTPUT"
          fi

      - uses: softprops/action-gh-release@v3
        with:
          tag_name: ${{ steps.release_meta.outputs.tag_name }}
          target_commitish: ${{ github.sha }}
          draft: false
          prerelease: ${{ steps.release_meta.outputs.prerelease }}
          make_latest: ${{ steps.release_meta.outputs.make_latest }}
          fail_on_unmatched_files: true
          files: |
            *.pdf
```

@rcalixte what do you think?

scripts/pdf/include.tex
0xAX, line 1: Can we move this and all related scripts below, pandoc configuration and so on to `./scripts/pdf` directory or something like this?

.github/workflows/gen-pdf.yaml
rcalixte: I see what you mean and I can work to implement this.

scripts/pdf/include.tex
rcalixte: We should be able to move things around. I'll include this in the update as well.

rcalixte: This works but it's kind of hacky in that the scripts still expect to be run from the subdirectory while the contents are a different level from the root. I opted for this behavior in the scripts since I think most folks running the scripts locally would do so from the directory itself. We can refactor if you think that assumption isn't fair.

.github/workflows/gen-pdf.yaml
rcalixte: The rebased push consolidates these changes. I put permissions at the top level but I'm flexible if you think it's better to be specified within the job's context. I don't expect this workflow would ever need more than the single job so it could be a moot point.

0xAX: Yes, I think it is pretty ok

scripts/pdf/include.tex
0xAX: @rcalixte yes, thank you. I think it is pretty ok to me.

0xAX:
@rcalixte thank you for the update. In general looks good to me! I will just test it a bit later today/tomorrow and if everything is good lets merge

scripts/pdf/include.tex
0xAX: I tested, works great.

To avoid the issue you mentioned, can you please just add:

```sh
# The pandoc-*.yaml configs and the paths inside them are relative to this
# script's directory, so always run from there regardless of where the
# script was invoked from.
cd "$(dirname "$0")" || exit 1
```

to the top of the pandoc.sh. It will make it work from both dirs, and from root dir and from the scripts dir.

0xAX:
@rcalixte Only little comment about pandoc.sh, can you please add it and let's merge!

scripts/pdf/include.tex
rcalixte: Great fix! Just when I thought I'd seen it all!

rcalixte: After discussing with a friend, I opted for `pushd`/`popd` over `cd` but I can revert if you want.

0xAX: > After discussing with a friend, I opted for pushd/popd over cd but I can revert if you want.

No no, it is cool!

0xAX (approved):

0xAX:
@rcalixte thank you a lot for your contribution! Merging.

